PR #1, opened 2020-04-13
doc: add shortened daxpy example to README

bd4:
Includes sample build command for the three platforms.


bd4:
> I'd also prefer to just write a*x + y instead of making a daxpy function for it, which, being templated, doesn't look all that simple.
That is a good point, I can save methods to a later example that talks about auto vs templated gtensor return values. Definitely more of an advanced topic, I'll update it in a future doc PR.

PR #11, opened 2020-05-19
add tests and missing assigner/equals

bd4:
Adding tests to prepare for SYCL commits, to make sure there are no regressions in CUDA and HIP.


bd4:
All tests pass and examples work in CUDA. My HIP/ROCm software stack got broken and I am unable to get it to work even on master, so not sure if there are any regressions there... going to try to find another machine to test with.

There is a new nvcc warning introduced by the new launch tests. Pretty sure this is an existing problem exposed by the new tests, but not 100% sure.

bd4:
Confirmed no regressions in HIP/ROCm support (and the new tests pass).

include/gtensor/gfunction.h
```diff
@@ -128,54 +128,69 @@ inline void calc_shape(S& shape, const E& e)
...
+struct gtensor_inner_types<gfunction<F, E1, E2>>
+{
+  using space_type = space_t<expr_space_type<E1>, expr_space_type<E2>>;
+  constexpr static size_type dimension = helper::calc_dimension<E1, E2>();
+
+  using value_type =
+    decltype(std::declval<F>()(std::declval<expr_value_type<E1>>(),
+                               std::declval<expr_value_type<E2>>()));
+  using reference = value_type;
+  using const_reference = value_type;
```
germasch, line 171: I don't think this is anything new in this PR, but let me just note that this seems to be hackiness that should be addressed eventually.

bd4: Created issue for it: https://github.com/wdmapp/gtensor/issues/12

PR #13, opened 2020-05-20
add CI for testing on CPU

bd4:


CMakeLists.txt
```diff
@@ -60,9 +60,34 @@ else()
...
+      add_subdirectory(${googletest_SOURCE_DIR} ${googletest_BINARY_DIR} EXCLUDE_FROM_ALL)
+    endif()
+    add_library(GTest::GTest INTERFACE IMPORTED)
+    target_include_directories(GTest::GTest INTERFACE "${googletest_SOURCE_DIR}/googletest/include")
+    target_link_libraries(GTest::GTest INTERFACE gtest)
+    add_library(GTest::Main INTERFACE IMPORTED)
+    target_link_libraries(GTest::Main INTERFACE gtest_main)
+  endif()
+else()
+  message(STATUS "build testing is OFF")
```
germasch, line 90: I think this is a bug, actually: This will happen if an external googletest is found, in which case tests will work, too.

germasch: Oops, actually that's not true.

bd4: I re-organized the ifs... CTest must be loaded for the BUILD_TESTING variable to be set correctly (defaults to on). If that is ON (hasn't been turned on explicitly), then looks for external Gtest, and if not found, downloads one to build internally.

PR #16, opened 2020-05-22
support multiple devices in cmake-config

bd4:


README.md
```diff
@@ -109,6 +141,46 @@ target_gtensor_sources(myapp PRIVATE src/myapp.cxx)
...
+git submodule add https://github.com/wdmapp/gtensor.git external/gtensor
+```
+
+In the application's `Makefile`:
+```make
+# gtensor configuration
+GTENSOR_DIR = external/gtensor
+GTENSOR_DEVICE = cuda
+
+GTENSOR_DEFINES = -DGTENSOR_HAVE_DEVICE -DGTENSOR_DEVICE_$(GTENSOR_DEVICE) -NDEBUG
```
germasch, line 163: I wouldn't put `-NDEBUG` here, as it's a generic macro not really specific to gtensor. And it's missing a `D`. (`-DNDEBUG`)

bd4: Only CUDA works without NDEBUG set - both HIP and SYCL fail to compile with the asserts.

I'll refactor the example so it doesn't use CXXFLAGS.

germasch: So maybe that's a good opportunity to replace the NDEBUG use by, e.g.., `GTENSOR_ENABLE_ASSERT`? (Name taken from xtensor).

germasch: So you could also merge it as-is, and refactor the asserts later.

bd4: good idea, created issue #17

PR #18, opened 2020-05-28
add gtensor_storage alternative to thrust device_vector

bd4:
Using thrust is still the default for both cuda and hip. This lays the groundwork for SYCL support, where a thrust implementation is not available.


bd4:
I think the gt::copy -> backend::memcpy still needs some refinement, need better boundary on when it changes from count based to bytes based. Maybe it should be backend::copy, with the low level device versions that wrap cudaMemcpy etc using the memcpy name and bytes, everything else is counts. Other than that I'm pretty happy with this.

include/gtensor/gtensor.h
```diff
@@ -84,14 +84,27 @@ inline gtensor<T, N, S>::gtensor(const shape_type& shape)
...
   // FIXME?! this kinda changes row-major list into transposed col-major array
   shape_type shape = helper::nd_initializer_list_shape<N>(il);
   base_type::resize(shape);
+#if defined(GTENSOR_HAVE_DEVICE) && !defined(GTENSOR_USE_THRUST)
+  if (std::is_same<S, space::device>::value) {
+    gtensor<T, N, space::host> host_temp(shape);
+    helper::nd_initializer_list_copy<N>(il, host_temp);
+    gt::memcpy<T, space::host, space::device>(base_type::data(),
+                                              host_temp.data(),
+                                              host_temp.size());
```
germasch, line 101: A couple of comments, but nothing that needs to be changed at this time:
* To me, it's more C++ to use `copy` rather than `memcpy`, and have the to/from reversed (you did follow the memcpy convention, so that's good. However, it looks like for the template args are the other way around?
If one used `copy`, it'd also be more consistent to use `begin`/`end`, but again it's consistent with the C memcpy.

Generally, though, it'd be nice to have a helper that doesn't need `data()`/`size()` or needs to know about the spaces, ie., just a `copy(host_temp, *this)`. And that function, at least kind of, does already exist. I think it's rather ugly, currently, as it handles all kinds of cases explicitly.

Finally, and that's not something I really have figured out whether I really like it or not, there's the consideration of being able to just say `*this = host_temp`.  Just mentioning it, definitely not now ;)

include/gtensor/gtensor.h
```diff
@@ -154,28 +167,28 @@ template <typename T, int N, typename S_from, typename S_to>
 void copy(const gtensor<T, N, S_from>& from, gtensor<T, N, S_to>& to)
 {
   assert(from.size() == to.size());
-  thrust::copy(from.data(), from.data() + from.size(), to.data());
+  gt::memcpy<T, S_from, S_to>(to.data(), from.data(), to.size());
```
germasch, line 170: Right, so you're obviously aware of `gt::copy`. I think it needs some cleaning up, but eventually it could be used above.

Also, there's no `gt::copy` and `gt::mempcy`, so I think the latter probably shouldn't be part of the public API, hence go into some `detail` or whatever namespace.

include/gtensor/gtensor_storage.h
```diff
@@ -0,0 +1,215 @@
...
+      if (capacity_ > 0) {
+        data_ = allocator_type::allocate(capacity_);
+      }
+    }
+    gtensor_storage() : gtensor_storage(0) {}
+
+    ~gtensor_storage() {
+      if (data_ != nullptr) {
+        allocator_type::deallocate(data_);
+      }
```
germasch, line 39: I think, but I could be wrong, that normally deallocate should handle a `nullptr` arg?

include/gtensor/gtensor_storage.h
```diff
@@ -0,0 +1,215 @@
...
+
+template <typename T>
+using host_storage = gtensor_storage<T, host_allocator<T>>;
+
+template <typename T, typename A>
+inline void gtensor_storage<T, A>::resize(gtensor_storage::size_type new_size) {
+  if (new_size == 0) {
+    if (data_ != nullptr)
+      allocator_type::deallocate(data_);
+    capacity_ = size_ = 0;
```
germasch, line 108: Is it standard behavior for `std::vector` etc to release memory if you resize them to 0? Not necessarily opposed to it, and I think it doesn't really happen in current use, but just wondering as I think it's best to mirror existing behavior where possible.

include/gtensor/gtensor_storage.h
```diff
@@ -0,0 +1,215 @@
...
+  } else if (capacity_ == 0) {
+    capacity_ = size_ = new_size;
+    data_ = allocator_type::allocate(capacity_);
+  } else if (new_size > capacity_) {
+    if (data_ != nullptr)
+      allocator_type::deallocate(data_);
+    capacity_ = size_ = new_size;
+    data_ = allocator_type::allocate(capacity_);
```
germasch, line 116: So this definitely looks like different semantics than the usual resize, which is not destructive for existing values. Obviously having to copy values that end up not being used is not good for performance, though I suppose in practice we don't really resize except for from 0. But anyway, I'm concerned about having something that's pretty similar to an existing type, but then has subtly different behavior in some ways. (Of course, one way it's different is that things values don't get initialized/constructed, but that at least is only specific thing.) So it may be worth considering to having this function named differently, though that's also a bit of a hassle because then one has to use some more trickery to use a `resize_discard` function when available, and `resize` otherwise...

include/gtensor/gtensor_storage.h
```diff
@@ -0,0 +1,215 @@
...
+  } else if (new_size < capacity_) {
+    // TODO: set shrink threshold?
+    size_ = new_size;
+  }
+}
+
+
```
germasch, line 123: Another minor comment: I feel like a single line of space is enough between related functions.

include/gtensor/macros.h
```diff
@@ -25,45 +25,45 @@
...
 
-#ifdef __CUDACC__
+#ifdef GTENSOR_DEVICE_CUDA
 
-#define gtLaunchKernel(kernelName, numblocks, numthreads, memperblock, streamId, ...)          \
-    do {                                                                                           \
-        kernelName<<<numblocks, numthreads, memperblock, streamId>>>(__VA_ARGS__);                 \
+#define gtLaunchKernel(kernelName, numblocks, numthreads, memperblock, streamId, ...) \
+    do { \
+        kernelName<<<numblocks, numthreads, memperblock, streamId>>>(__VA_ARGS__); \
```
germasch, line 32: Do you run clang-format on things? I propose doing that, though I'm not always good at doing that myself.

That would also be a nice thing to add to the CI, make sure a PR is properly formatted. I do have a script to do that, basically taken from adios2, though it's for travis CI.

include/gtensor/gtensor.h
bd4: Good point, regardless of which convention it follows it should be moved to somewhere like `gt::backend`. This is basically the internal implementation of `gt::copy` when thrust is not being used. There are necessarily different layers to the copy interface. There are wrappers around the raw device routines - device_memcpy_{dd,hh,dh,hd} (must be implemented when adding a new backend), the alias versions used via the allocator interface in `gtensor_storage` (wrap dd/hh depending on allocator type), and the versions used to implement `gt::copy`. The `gt::copy` layer knows the space because everything is captured in the CRTP expression types. The lower level layers don't / can't have this, so they must be explicit. SYCL usm allocated pointers containing the location info, but CUDA and HIP do not (or at least the memcpy interface doesn't rely on it - I think later versions of cuda do have a enum for inferring the direction). I'm inclined to expose the low level versions for writing tests and for when it's needed, rather than trying to hide it. Definitely not for 'public' use though.

I prefer `gt::backend::copy(from, to, size)` over `gt::backend::copy(from, from+size, size)`. Seems like a pointless extra pointer / iterator arithmetic. This is arguably also more internally consistent with `gt::copy(from, to)`, but not consistent with `std::copy`...

bd4: I pushed a refactored version that does `gt::backend::copy(from, to, count)` and `gt::backend::device_copy_*` with same convention. The memcpy's now exist only inside the implementations. I think it's a lot cleaner now.

include/gtensor/macros.h
bd4: I'd prefer to do the clang-format after as part of #19 to simplify rebase/merge. I can also add CI for it.

PR #19, opened 2020-05-29
add SYCL support

bd4:


bd4:
Note: this should only be merged after #18

PR #22, opened 2020-06-03
Pr/test complex

bd4:


include/gtensor/complex.h
```diff
@@ -11,14 +11,25 @@
...
 #endif
 
+template <typename T>
+struct is_complex : public std::false_type
+{};
+
+template <typename T>
+struct is_complex<complex<T>> : public std::true_type
+{};
+
```
germasch, line 32: Out of curiosity, what's this needed for?

bd4: I was experimenting with doing `enable_if_t<is_complex<expr_value_type<E>>::value>` to make a thrust::complex friendly version of the nd initializer list stuff, but that doesn't seem to work. Might be useful in the future though? I could remove it.

bd4:
The HIP check is failing because of ROCm 3.5.0 update, which has been automatically used by the CI. Fix in other PR which could be merged first.

include/gtensor/complex.h
germasch: It's small and doesn't hurt. If you think you'll need it in the future, you can keep it, though generally I prefer waiting to add something until there's an actual need. I haven't look at the initializer stuff at all...

germasch:
Yeah, I guess you can merge the other one first and then rebase this one.

PR #28, opened 2020-06-25
Pr/fix view copy

bd4:
adds gview move ops, gscalar store decayed type, lots of tests

### DIFF
--- a/include/gtensor/gscalar.h
+++ b/include/gtensor/gscalar.h
@@ -35,9 +35,9 @@ private:
 };
 
 template <typename T>
-gscalar<T> scalar(T t)
+gscalar<T> scalar(T&& t)
 {
-  return gscalar<T>(t);
+  return gscalar<T>(std::forward<T>(t));
 }
 
 } // namespace gt
--- a/include/gtensor/gscalar.h
+++ b/include/gtensor/gscalar.h
@@ -18,7 +18,7 @@ public:
 
   constexpr static size_type dimension() { return 0; }
 
-  gscalar(T&& value) : value_(value) {}
+  gscalar(T value) : value_(value) {}
 
   shape_type shape() const { return {}; }
 
@@ -28,19 +28,16 @@ public:
     return value_;
   }
 
-  gscalar<value_type> to_kernel() const
-  {
-    return gscalar<value_type>(value_type(value_));
-  }
+  gscalar<value_type> to_kernel() const { return gscalar<value_type>(value_); }
 
 private:
-  T value_; // FIXME? would be much less fragile to never store reference
+  value_type value_;
 };
 
 template <typename T>
-gscalar<T> scalar(T&& t)
+gscalar<T> scalar(T t)
 {
-  return gscalar<T>(std::forward<T>(t));
+  return gscalar<T>(t);
 }
 
 } // namespace gt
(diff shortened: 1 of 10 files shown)


include/gtensor/gscalar.h
```diff
@@ -28,19 +28,16 @@ class gscalar : public expression<gscalar<T>>
...
 
-  gscalar<value_type> to_kernel() const
-  {
-    return gscalar<value_type>(value_type(value_));
-  }
+  gscalar<value_type> to_kernel() const { return gscalar<value_type>(value_); }
 
 private:
-  T value_; // FIXME? would be much less fragile to never store reference
+  value_type value_;
```
germasch, line 34: I don't mind taking this for now. However, I'll say that in the long run I hope to get back to some more extensive discussion about the benefits/drawbacks (The way I see it right now, the benefit is it avoids a certain class of subtle bugs, but on the other hand it makes the semantics different from  other expressions which do take references.)

In the end, the outcome could also be to never take references in other expressions, either, with the exception of using a `gtensor_view` where currently a reference to a `gtensor` is kept, ie., only avoid copying of actual underlying data, but otherwise just replicate the object.

include/gtensor/gscalar.h
```diff
@@ -28,19 +28,16 @@ class gscalar : public expression<gscalar<T>>
...
 };
 
 template <typename T>
-gscalar<T> scalar(T&& t)
+gscalar<T> scalar(T t)
 {
-  return gscalar<T>(std::forward<T>(t));
+  return gscalar<T>(t);
 }
```
germasch, line 41: I'm not sure i'd change this -- it's just basically a way to derive template args for calling the constructor, and that's something where perfect forwarding is always appropriate (and wouldn't make a difference here.)

bd4: Ok - I see several places where the template helper uses perfect forwarding, and the constructor takes a universal reference, but it does not use `std::forward` while constructing it's member expression (e.g. gview_adapter does this). Seems like that should also forward?

include/gtensor/gscalar.h
bd4: One place where storing a reference breaks down is when the reference is to a function parameter. This is much more likely in to happen with scalars than for gtensor objects. Copying gtensors is expensive, so generally they are passed by reference and have a lifetime that exists beyond the function call. Scalars on the other hand are inexpensive to pass by value and it is common practice. This is why I think it isn't so strange to special case them.

Using `to_kernel` on sub-expressions before storing in expression members is another option - that would have the effect of converting to gtensor_view.

include/gtensor/gscalar.h
germasch: This?

```cxx
  gview_adaptor(E&& e) : e_(e), strides_(calc_strides(e.shape())) {}
```

Yes, I believe this should have used `std::forward`. It's not fatal, if a reference is passed, it makes no difference, but if a rvalue is passed it leads to a copy, rather than a move, which should still work right, but not as intended.

I think at some point the whole gtensor thing should probably go through a thorough code review...

include/gtensor/gscalar.h
germasch: Right -- I don't have a set opinion on which way is better, so I don't mind merging this. It'd be nice to get rid of `to_kernel()` altogether  -- kinda just always do it. But I don't think that's possible, since sometimes expressions really do own a `gtensor`, so then one can't just convert it to a `gtensor_view`. I guess reference counting is an alternative, may be worth considering at some point.

PR #29, opened 2020-06-28
sarray: add data() methods

germasch:
This is just mimicking standard container behavior.


bd4:
> This is just mimicking standard container behavior.

So mainly for interface consistency?

germasch:
I actually needed it at some point debugging something with the ongoing gene Poisson remapping work. I don't remember the details right now, it's been a while...

PR #31, opened 2020-08-13
Pr/add synchronize

bd4:
Adds `gt::synchronize` device call, mapped to `cuda(hip)StreamSynchronize(0)` and `q.wait()` for SYCL. The eventual goal is to be able to write GPU vendor independent code using gtensor without having to do any vendor specific code hacks in client code, and synchronize is a pretty common requirement. In particular this is used in GENE.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13...3.17.2 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.13...3.18 FATAL_ERROR)
 project(gtensor
         VERSION 0.01
         LANGUAGES CXX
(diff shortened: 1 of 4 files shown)


bd4:
Needs a dummy host implementation.

bd4:
Note that the SYCL tests fail because the test image always pulls the latest intel dpcpp version. Currently beta8, which introduces changes that break the existing code, unrelated to this PR. This is a potential improvement for the CI, but Intel does not provide version locked repositories, so would need to build and host our own docker images on docker hub to make that work, so doesn't seem worth it atm. For now I think keeping the SYCL tests as non-required for merge is fine, since it's evolving very quickly.

germasch:
Ok, so I guess you had already done that, so good to go.

PR #32, opened 2020-09-04
fix for rocthrust 3.7

bd4:
rocthrust has an older verison, but uses the newer API.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -178,14 +178,9 @@ jobs:
 
   build-hip:
     runs-on: ubuntu-latest
-    #container: rocm/dev-ubuntu-18.04:3.3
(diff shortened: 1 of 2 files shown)


bd4:
Note that HIP tests keep failing because of https://github.com/RadeonOpenCompute/ROCm/issues/1160. I'm going to switch to a vanilla debian or ubuntu docker image and manually add the repo and key, then install all ROCm from scratch. Hopefully that will be less fragile.

germasch (approved):
Looks good to me (but looking at it is all I've done...)

bd4:
Ignoring SYCL failure for now, ROCm is working again after CI changes to use vanilla ubuntu and manually set up ROCm repo and install latest, rather than trying to use their fragile docker hub image.

PR #33, opened 2020-09-16
fix calc_shape for nested expressions with unary sub-expressions

bd4:


bd4:
I landed on `gt_empty_expr` for the new name, more descriptive than `gt_empty` or `gt_internal_empty` or `internal::empty`.

germasch (approved):

PR #35, opened 2020-09-19
Pr/sycl fixes

bd4:
Tests and daxpy example run with Intel DPCPP beta09 on CPU, and on GPU on certain Argonne test machines (but unable to get GPU working on my laptop). In any case this should not affect non SYCL backends, and moves things much closer to 2020 provisional spec compliance.

The biggest missing feature here is how to handle explicit kernel launch. This is not exercised well yet by the test cases and the one GPU example.

### DIFF
--- a/.github/workflows/sycl.yml
+++ b/.github/workflows/sycl.yml
@@ -8,10 +8,10 @@ jobs:
     #container: intel/oneapi-basekit:devel-ubuntu18.04
     container: ubuntu:18.04
     env:
-      CXX: /opt/intel/inteloneapi/compiler/latest/linux/bin/dpcpp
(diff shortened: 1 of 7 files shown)


germasch (approved):
LGTM.

I should actually learn to basics of sycl some time, I guess. So the kernel name template parameter is essentially just that, just a name without any intrinsic meaning of the type passed? I think it's kinda cool to have the ability to name a kernel, for debugging / profiling, if that's what it's for.

bd4:
Yes that's all it's for, you don't even have to pre-declare it. Intel has an extension that allows auto-gen of names, but I haven't been able to get that to work in the way I expect yet, and in any case the auto-gen names will likely not be so friendly.

PR #36, opened 2020-09-21
Pr/device management

bd4:


bd4:
I think a the cuda kernels in GENE will work in HIP just by changing the special launch syntax to use gtLaunchKernel. This is actually already present since gtensor.h includes device_backend.h which includes macros.h. I'm thinking gtensor/capi.h should also include macros.h, I think it's all C safe now.

bd4:
Using them is a hack since they won't work with SYCL. But sometimes doing a hack first makes sense... gives us some time as the SYCL implementations mature, and for GENE branches that already have gtensor ports of kernels to be merged.

include/gtensor/device_backend.h
```diff
@@ -81,6 +114,29 @@ struct device_allocator
   }
 };
 
+template <typename T>
+struct managed_allocator
+{
+  static T* allocate(int count)
```
germasch, line 120: This should probably be `size_t` rather than `int`. My general take on this is that if one allocates a m x n x l array, it's usually okay for the individual dims to be `int` (though one can definitely argue about that, too). But allocating something > 4GB (or 4e9 x sizeof(float)) is something that can actually happen.

include/gtensor/device_backend.h
```diff
@@ -81,6 +114,29 @@ struct device_allocator
...
+
+  static void deallocate(T* p)
+  {
+    if (p != nullptr) {
+      gtGpuCheck(cudaFree(p));
+    }
+  }
+
+  static void copy(const T* src, T* dst, std::size_t bytes)
+  {
```
germasch, line 135: And so while this is `size_t`, I think it should be the same type used above, because I also think it should be a count, rather than bytes. Since this is templated by `T`, I think the API should be consistently in terms of count of elements.

src/cgtensor.cxx
```diff
@@ -0,0 +1,56 @@
...
+extern "C" void gt_synchronize()
+{
+  gt::synchronize();
+}
+
+#ifdef GTENSOR_HAVE_DEVICE
+
+#ifndef GTENSOR_DEVICE_SYCL
+
+extern "C" int gt_backend_device_get_count()
```
germasch, line 14: I think if you declare the function `extern "C"` in the header, you don't have to repeat that here.

include/gtensor/capi.h
```diff
@@ -0,0 +1,31 @@
...
+#define GTENSOR_CAPI_H
+
+/**
+ * Expose a limited subset of gtensor API for calling from Fortran.
+ *
+ * The C++ namespaces are adapted to underscores when possible. For
+ * allocation, the calls are simplified to exclude the allocator type.
+ */
+
+extern "C" void gt_synchronize();
```
germasch, line 11: I think you'll find that you can't include this from C, since C doesn't understand `extern "C"` ;)

I like going the `{BEGIN.END}_C_DECLS` (macros) way.

include/gtensor/device_backend.h
```diff
@@ -158,6 +247,29 @@ struct device_allocator
...
+  }
+
+  static void deallocate(T* p)
+  {
+    if (p != nullptr) {
+      gtGpuCheck(hipFree(p));
+    }
+  }
+
+  static void copy(const T* src, T* dst, std::size_t count)
```
germasch, line 267: When the arg is called `count`, it's yet more misleading. I haven't looked at the existing allocators, but I'd argue the same there, if their interface is like this, too.

germasch (changes requested):
Looks pretty good, but I do have a couple of comments.

include/gtensor/device_backend.h
bd4: ok using gt::size_type everywhere. It's actually all counts, I missed one of the 'bytes' names when I made the change and then copied it when I made the managed allocator.

include/gtensor/device_backend.h
bd4: it is truly a count (number of T's)

include/gtensor/device_backend.h
germasch: In the C routines for Fortran interfacing, you should use a standard `size_t` (a.k.a. `C_SIZE_T` from Fortran) to avoid potential trouble there.

germasch:
github says the branch is out of date w.r.t. master. My preference on this is to rebase on top of master and force push before merging, unless that has the risk of subtly breaking stuff along the way.

include/gtensor/capi.h
```diff
@@ -14,24 +14,32 @@
...
+uint32_t gt_backend_device_get_vendor_id(int device_id);
 
 #endif // not GTENSOR_DEVICE_SYCL
 
-extern "C" void* gt_backend_device_allocate(int nbytes);
-extern "C" void* gt_backend_managed_allocate(int nbytes);
-extern "C" void gt_backend_device_deallocate(void* p);
-extern "C" void gt_backend_managed_deallocate(void* p);
+void* gt_backend_device_allocate(size_t nbytes);
+void* gt_backend_managed_allocate(size_t nbytes);
```
germasch, line 35: Ok, it already has `size_t`, so great ;)

germasch (approved):

PR #37, opened 2020-09-24
More clib

bd4:


include/gtensor/complex_ops.h
```diff
@@ -0,0 +1,54 @@
...
+namespace gt
+{
+
+#if defined(GTENSOR_DEVICE_CUDA) || defined(GTENSOR_USE_THRUST)
+
+template <typename T>
+GT_INLINE T norm(const complex<T>& a)
+{
+  return thrust::norm(a);
+}
```
germasch, line 16: Since we talked about ADL the other day: Is this really needed? Ie., does `norm(a)` not just work?

include/gtensor/complex_ops.h
```diff
@@ -0,0 +1,54 @@
...
+{
+  return thrust::conj(a);
+}
+
+template <typename T>
+GT_INLINE T norm(thrust::device_reference<complex<T>> a)
+{
+  return thrust::norm(thrust::raw_reference_cast(a));
+}
+
```
germasch, line 29: This one should definitely be needed, though.

In the longer term, I'm wondering whether to maybe get rid of that whole `device_reference` thing, which is trying to make device values accessible on the host like normal, which kinda mostly works, but then still leads to problems every so often.

bd4: I've been thinking about that too - the thrust wrapper could be removed before returning from `operator()` etc, then it would be more consistent with the non-thrust backend(s). I've already updated the unit tests to not rely on the host access to device_reference feature.

include/gtensor/complex_ops.h
bd4: That is a good question. The issue here is that complex is actually in gt namespace, via a using statement, even though the aliased from type is in thrust (or std for HIP nothrust/SYCL). Based on a quick test, I think ADL is not smart enough to look at the aliased from type, build-hip-nothrust broke when I removed the wrappers in gt.

germasch: Ok, makes sense.

include/gtensor/complex_ops.h
germasch: Yeah -- and eventually, maybe we can get rid of the whole dependence on thrust? I don't know, but there's a whole bunch of ifdefs etc now, it wouldn't hurt to have fewer cases to support. But in any case, what you're suggesting is a good initial step (probably -- I haven't thought about it too hard).

germasch (approved):

include/gtensor/complex_ops.h
bd4: The annoying thing is that CUDA, last I checked, does not work with std::complex in device code. That is why there is ifdef hackery to use thrust::complex and thrust_ext.h if GTENSOR_DEVICE_CUDA or GTENSOR_USE_THRUST. So even when use thrust is off for cuda, it uses thrust::complex (but still uses the non-thrust array backend).

germasch: Are we setting `-expt-relaxed-constexpr`? This might make it work (I haven't checked) -- if the `std::complex` stuff is `constexpr` (which I think it would make sense, but I don't know whether it is...)

PR #38, opened 2020-09-25
Pr/sycl launch

### LINKED ISSUE #30: sycl: update for 2020 provisional standard
One important change is that mutable lambdas are no longer allowed, so const gtensor/view refs need to allow element modification.

bd4:
Update launch to work with Intel dpcpp beta09, which is at least a partial implementation of SYCL 2020 provisional. Closes #30. See also discussion at https://github.com/intel/llvm/issues/2538


bd4:
Note that this passes the (manually run) SYCL CI tests.

germasch (approved):
LGTM.

In the long run, we may really have to fix gtensor, though...

PR #42, opened 2020-10-12
fix: call sync in caching deallocate

germasch:
Usually, cudaFree() is synchronizing, and that needs to be done in the
caching allocator, too, since it's not a good idea to free memory while
it's still being used in a concurrently executing kernel.

### DIFF
--- a/include/gtensor/space.h
+++ b/include/gtensor/space.h
@@ -65,9 +65,7 @@ struct caching_allocator : A
 
   void deallocate(pointer p, size_type cnt)
   {
-#ifdef GTENSOR_HAVE_DEVICE
-    cudaCheck(cudaDeviceSynchronize());
-#endif
+    gt::synchronize();
     auto it = allocated_.find(p);
     assert(it != allocated_.end());
     free_.emplace(std::make_pair(it->second, p));
--- a/include/gtensor/space.h
+++ b/include/gtensor/space.h
@@ -65,6 +65,9 @@ struct caching_allocator : A
 
   void deallocate(pointer p, size_type cnt)
   {
+#ifdef GTENSOR_HAVE_DEVICE
+    cudaCheck(cudaDeviceSynchronize());
+#endif
     auto it = allocated_.find(p);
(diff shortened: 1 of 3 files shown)


bd4 (commented):
change to `#ifdef GTENSOR_DEVICE_CUDA` and `gtGpuCheck` instead of `cudaCheck`

bd4:
> change to `#ifdef GTENSOR_DEVICE_CUDA` and `gtGpuCheck` instead of `cudaCheck`
Actually can just use gt::synchronize() here and no ifdef. It's a no-op on host, and does the right thing for various backends. Even though currently caching allocator is only instantiated for GTENSOR_USE_THRUST.

include/gtensor/space.h
```diff
@@ -65,6 +65,9 @@ struct caching_allocator : A
 
   void deallocate(pointer p, size_type cnt)
   {
+#ifdef GTENSOR_HAVE_DEVICE
+    cudaCheck(cudaDeviceSynchronize());
+#endif
```
bd4, line 70: Should be able to replace all three lines with `gt::synchronize()`, if include order allows.

germasch: Sounds like a good plan ;)

bd4: Ok so that didn't work. gt::synchronize just calls gt::backend::device_synchronize, defined in device_backend.h. That header has no gtensor deps, so it should work. I'm not sure how I feel about the dual definitions, but maybe is ok to use the backend one internally. You will need to wrap with GTENSOR_HAVE_DEVICE ins that case

germasch: I moved `gt::synchronize()` into `device_backend()` now, which isn't the prettiest solution, but let's see how it fares.

I think `gt::synchronize()` is probably the only wrapper that replicates a backend function -- at some point in the future, maybe it makes sense to either wrap them all or not at all...

bd4 (approved):
Looks good now, thanks!

PR #46, opened 2020-11-13
add gpublas and gpufft libraries

bd4:


CMakeLists.txt
```diff
@@ -12,13 +12,30 @@ set(GTENSOR_TEST_DEBUG OFF CACHE BOOL "Turn on debug printing for unit tests")
 set(GTENSOR_BUILD_DEVICES "" CACHE STRING "List of device types 'host', 'cuda', 'hip', and 'sycl' (semicolon separated)")
 set(GTENSOR_BUILD_EXAMPLES OFF CACHE BOOL "Build example programs")
 set(GTENSOR_BUILD_CLIB OFF CACHE BOOL "build cgtensor.a")
+set(GTENSOR_BUILD_BLAS OFF CACHE BOOL "build gpublas.a")
+set(GTENSOR_BUILD_FFT  OFF CACHE BOOL "build gpufft.a")
```
germasch, line 16: It's kinda minor, but if you build a static library, it'd presumably be called `libcgtensor.a` etc. But I also think it's not important to think about static vs shared here, so I'd just say `Build libcgtensor` etc. (With a capital B for consistency).

Though to add another thought, they're not just being built, but also installed (I suppose), so maybe it'd be better to say `GTENSOR_ENABLE_CLIB` etc in the first place. (For the examples, "build" is good, but those choices do go beyond just building.)

CMakeLists.txt
```diff
@@ -12,13 +12,30 @@ set(GTENSOR_TEST_DEBUG OFF CACHE BOOL "Turn on debug printing for unit tests")
...
+    STRING "comma separated list of AMD gpu architectures to build for")
 
 # SYCL specific configuration
-set(GTENSOR_DEVICE_SYCL_SELECTOR "default" CACHE STRING "Use specified sycl device selector ('default', 'gpu', 'cpu',  or 'host')")
-set(GTENSOR_DEVICE_SYCL_INTEL ON CACHE BOOL "Using Intel OneAPI SYCL implementation")
-set(DPCPP_ROOT "/opt/intel/inteloneapi/compiler/latest/linux" CACHE STRING "Path to DPCPP / Intel OneAPI root")
-set(DPCPP_LIBDIR "${DPCPP_ROOT}/lib" CACHE STRING "Path to DPCPP compiler lib dir")
-set(DPCPP_INCDIR "${DPCPP_ROOT}/include/sycl" CACHE STRING "Path to DPCPP include dir")
+set(GTENSOR_DEVICE_SYCL_SELECTOR "default" CACHE STRING
+    "Use specified sycl device selector ('default', 'gpu', 'cpu',  or 'host')")
```
germasch, line 27: Maybe it's time to start using `cmake-format`? ;)

CMakeLists.txt
```diff
@@ -88,7 +105,7 @@ if ("hip" IN_LIST GTENSOR_BUILD_DEVICES)
     message(FATAL_ERROR "For GTENSOR_BUILD_DEVICES=hip, 'hcc' or 'hipcc' must be used as C++ compiler.")
   endif()
 
-  find_package(HIP REQUIRED CONFIG PATHS "/opt/rocm/hip")
+  find_package(HIP REQUIRED CONFIG PATHS "${HIP_PATH}")
```
germasch, line 108: Usually, one should be able to set `HIP_ROOT` and cmake will automatically look there? Ideally the consumer (gtensor) shouldn't need to define its own set of search variables, but the ones from the find module/config should be usable directly.

CMakeLists.txt
```diff
@@ -98,11 +115,46 @@ if ("hip" IN_LIST GTENSOR_BUILD_DEVICES)
...
-    target_link_libraries(gtensor_hip INTERFACE roc::rocthrust)
+
+    # Manually define ROCm libraries. Using ROCm's cmake support pollutes
+    # flags that appear in Fortran builds, and breaks some mixed language
+    # applications (GENE in particular).
+    add_library(rocthrust INTERFACE IMPORTED)
+    target_include_directories(rocthrust INTERFACE
+                               "${ROCM_PATH}/rocprim/include")
+    target_include_directories(rocthrust INTERFACE
+                               "${ROCM_PATH}/rocthrust/include")
```
germasch, line 126: So one part of course would be to try to get AMD to fix their stuff so that it's usable. If that's not happening anytime soon, I think it might make sense to write a little find module for `rocthrust` and other stuff, that either does what you do above, or uses the config and then fixes it up.

CMakeLists.txt
```diff
@@ -209,6 +272,34 @@ if (GTENSOR_BUILD_CLIB)
...
+
+  if (${GTENSOR_DEVICE} STREQUAL "cuda")
+    target_link_libraries(gpublas "cublas")
+  elseif (${GTENSOR_DEVICE} STREQUAL "hip")
+    target_link_libraries(gpublas rocblas rocsolver)
+  elseif (${GTENSOR_DEVICE} STREQUAL "sycl")
+    target_link_libraries(gpublas oneapi_mkl_sycl)
+  endif()
+endif()
+
```
germasch, line 288: This may be worth a new subdirectory, for blas, and for fft?

include/gtensor/gpufft.h
```diff
@@ -0,0 +1,90 @@
...
+#include "CL/sycl.hpp"
+//#include "oneapi/mkl.hpp"
+#include "oneapi/mkl/dfti.hpp"
+#include "mkl.h"
+
+typedef enum gpufft_transform_enum
+{
+  GPUFFT_R2C = 0x2a, // Real to complex (interleaved)
+  GPUFFT_C2R = 0x2c, // Complex (interleaved) to real
+  GPUFFT_C2C = 0x29, // Complex to complex (interleaved)
```
germasch, line 49: Where do these magic numbers come from? It'd be nice if that could be avoided.

src/gpublas.cxx
```diff
@@ -0,0 +1,314 @@
...
+{
+#ifdef GTENSOR_DEVICE_CUDA
+  gtGpuCheck((cudaError_t)cublasCreate(&handle));
+#elif defined(GTENSOR_DEVICE_HIP)
+  gtGpuCheck((hipError_t)rocblas_create_handle(&handle));
+#elif defined(GTENSOR_DEVICE_SYCL)
+  handle = &gt::backend::sycl::get_queue();
+#endif
+}
+
```
germasch, line 18: In the long run, I think it'd be desirable to avoid all these ifdefs in this file (ie., splitting things into different source files).

germasch (approved):
I only have some pretty minor comments.

include/gtensor/gpufft.h
bd4: they come from cuFFT / hipFFT. agreed they are ugly. The idea here is to support a common subset of all three APIs, but there are pros and cons to trying to be similar to the original, e.g. cuFFT. If it's similar but slightly different, that could be really confusing - get the baggage of the API design, plus confusion. So may be better to think through how to present the common subset in the most sensible way. I think that's out of scope for this PR though..

CMakeLists.txt
bd4: Ya I think that's a good idea, especially as the API expands and refactored to use more source files. I'll amend the PR for this.

CMakeLists.txt
bd4: I agree in theory but I could easily sink several days into that... so short term I am thinking to just move my interface library declarations to a separate cmake file, so there is not so much ugliness in the main file. Will amend PR accordingly.

CMakeLists.txt
bd4: I have yet to find a coherent example or description of how cmake is intended to be used with HIP. This approach comes from looking at HIP libraries. I'll take a quick look at hipMAGMA and hip Kokkos, but don't want to sink a lot of time into it now. It's all awful, at least with manual interface lib definitions it's easy to debug and get working on different machines.

CMakeLists.txt
bd4: I was thinking that would be nice to have - is there a canonical one? There is a python package in PyPI.

CMakeLists.txt
bd4: Good idea, I'll make the changes for this PR.

src/gpublas.cxx
bd4: Agreed. I am not sure it's worth making the change now though, iterating on the API design first makes sense because it all probably should change a bit.

CMakeLists.txt
bd4: Now that I am thinking about this more, I'm not sure what the best way to do it is. Having extra include directories just complicates things, e.g. `gpublas/{src,include}`.

CMakeLists.txt
germasch: I think that python package should be fine, though I don't know -- I've never actually used it.

CMakeLists.txt
germasch: I guess my main point is that one should be able to skip HIP_PATH. If using spack, it'll set CMAKE_PREFIX_PATH for hip (in theory, anyway), so it'll find it from there. If setting things up manually, one can use `-DHIP_ROOT=...` if it's not found. Again, in theory, haven't tried any of these, but that's how it works (or should work) for everything else.

CMakeLists.txt
germasch: Sounds good.

CMakeLists.txt
germasch: Yeah, that's true, and I didn't think too much about it. The header files should all be under one `include/gtensor/`. One could add another level of subdirectories, but that's not really worth it if there's only a single header file, which I think is what it ends up being.  And if the includes don't get split up, it doesn't make much sense for the sources, either.

In the future, I could envision some kind of `#include <gtensor/blas/cuda.h>` or something, but only used internally when including `<gtensor/gtblas.h>` and when the backend is cuda. But let's cross that bridge when/if we get there...

src/gpublas.cxx
germasch: Fine w/me.

CMakeLists.txt
bd4: There aren't actually any ROCm or HIP spack packages. There is a hipsycl spack package, and they didn't even try to support ROCm, it's turned off, with CUDA support optional as a variant :P.

Setting CMAKE_PREFIX_PATH manually works I think. But which subdir of rocm install is that supposed to be? I think it's lib, but arguably this is even less friendly than manually setting HIP_PATH. And actually for anyone doing this on a dev machine, it's alwasy /opt/rocm and /opt/rocm/hip for ROCM/HIP paths, so they wouldn't need to change the defaults.

bd4: Looks like improving cmake support is being worked on: https://github.com/ROCm-Developer-Tools/HIP/issues/2158

I'm inclined to see how this evolves upstream and not spend a lot of time on it now. There are similar issues with oneAPI MKL, which I also hacked manual INTERFACE targets for.

PR #48, opened 2020-11-25
Pr/fft z2z

bd4:
Add complex-complex and single precision versions of all exec routines.


bd4:
This builds and runs in SYCL as well. There are some strange issues where having both double and single precision tests runs of r2c or c2r in same executable fails - I think this is a SYCL bug though, and not an issue with the library wrapper.

germasch (approved):
LGTM

PR #50, opened 2020-12-11
ci: add clang-format checker

germasch:


germasch:
Oops, got to realize there's already a clang-format checker, so this is pointless...

PR #51, opened 2021-01-21
add ij_deriv benchmark

bd4:
Not build by default, needs -DGTENSOR_BUILD_BENCHMARKS=ON


bd4:
A lot more can be added, like a Fortran version, but I think this is a reasonable starting point.

germasch (approved):
LGTM

PR #52, opened 2021-01-31
add sum/min/max reductions

bd4:


bd4:
For non-thrust, we could use native primitive libraries, e.g. cub for cuda, rocPRIM for HIP... but then HIP still needs an extra library (rocPRIM is not part of main lib). On the other hand, using thrust for reductions does not have to go together with using it as an array backend.

bd4:
One thing I am still considering, is how to refactor the GTENSOR_USE_THRUST cmake option (and the corresponding C preprocessor define). The simplest option is to just redefine what this means without a rename - basically now it means use thrust vectors for backing store, and thrust is always used for reductions regardless. A rename of the cmake variable is relatively easy and noninvasive - changing the C define is annoying, because all of the GNU make code that sets it manually will have to be updated.

include/gtensor/device_copy.h
```diff
@@ -114,7 +114,7 @@ namespace backend
 {
 
 template <typename T, typename S_from, typename S_to>
-inline void copy(T* src, const T* dest, std::size_t count)
+inline void copy(const T* src, T* dest, std::size_t count)
```
germasch, line 117: I was going to ask how this used to work, but I guess the subsequent casts explain it.... ;)

include/gtensor/expression.h
```diff
@@ -81,37 +81,37 @@ using to_kernel_t =
...
-auto index_expression(E expr, shape_type<5> idx)
+GT_INLINE decltype(auto) index_expression(E expr, shape_type<5> idx)
 {
   return expr(idx[0], idx[1], idx[2], idx[3], idx[4]);
 }
 
 template <typename E>
-auto index_expression(E expr, shape_type<6> idx)
+GT_INLINE decltype(auto) index_expression(E expr, shape_type<6> idx)
 {
```
germasch, line 115: I guess doing this in a generic way would be kinda nice, but IIRC, also not so straightforward... (`std::index_sequence`, I thinkj)

include/gtensor/meta.h
```diff
@@ -0,0 +1,17 @@
+#ifndef GTENSOR_META_H
+#define GTENSOR_META_H
```
germasch, line 2: Just as a small future todo: I like the `meta` namespace, I think there's likely other stuff which could be moved there, not that I remember anything specific right now...

include/gtensor/sarray.h
```diff
@@ -134,6 +137,33 @@ GT_INLINE T* sarray<T, N>::end()
   return data_ + N;
 }
 
+template <typename T, std::size_t N>
+GT_INLINE sarray<T, N + 1> sarray<T, N>::insert(std::size_t i, T value) const
```
germasch, line 141: The signature is different from the usual container `insert`, which is based on iterators. Not sure, though, if consistency doesn't make things too ugly. (Ie., one would have to do something like `shape.insert(shape.being() + i, val)`.

include/gtensor/sarray.h
```diff
@@ -134,6 +137,33 @@ GT_INLINE T* sarray<T, N>::end()
...
+  }
+  out[i] = value;
+  for (int j = i; j < N; j++) {
+    out[j + 1] = data_[j];
+  }
+  return out;
+}
+
+template <typename T, std::size_t N>
+GT_INLINE sarray<T, N - 1> sarray<T, N>::remove(std::size_t i) const
```
germasch, line 155: Similarly, usually this is `erase` but takes an iterator. But I'm also just realizing that it's quite different anyway, in that the usual ops happen in place, which isn't even possible for `sarray`.

So I think this is probably fine, though an alternative might be to make these freestanding function instead, that would be another way to make clear that they have different semantics, and are out-of-place.

tests/test_expression.cxx
```diff
@@ -251,6 +251,55 @@ TEST(expression, shape_second)
...
+  auto strides = a.strides();
+
+  auto k_a = a.to_kernel();
+  auto k_b = b.to_kernel();
+
+  gt::launch<1, S>(
+    linear_shape, GT_LAMBDA(int i) {
+      auto idx = unravel(i, strides);
+      index_expression(k_a, idx) =
+        index_expression(k_a, idx) + 2 * index_expression(k_b, idx);
```
germasch, line 271: So I guess I'm wondering why `index_expression` is done as a helper function, and whether that's necessary. I'd say it doesn't look particularly pretty ;)

One option would be to `operator[]`: `k_a[idx] = k_a[idx] + 2 * k_b[idx]`. I also think that I remember that xtensor has a way to pass the index as a tuple or something, not sure whether to `operator()` or `operator[]` or maybe it's different anyway....

germasch (approved):
LGTM. I do have some comments, but none are showstoppers.

tests/test_expression.cxx
bd4: Definitely ugly... I guess it comes down to whether we think this is useful to expose as part of the public API, or if we want to keep in for internal use (in which case index_expression should be moved to an inner namespace...).

There are two potential operations here - taking a sarray and/or tuple argument (which is what index_expresison does), and taking a linear index (which is actually what is used here and for SYCL's N dimension kernels). I think linear indexing is pretty useful. There is some subtlety here too - the tuple index must be based on the expression shape, not the internally stored strides which may be different for a gview, and are already taken care of by the gview (so using the internal strides effectively map twice and break things).

include/gtensor/sarray.h
bd4: I do feel weird about the names, since they seem to imply in place. Making them helper functions is a reasonable idea, but also makes it harder for people to find them, and I think they may be generally useful. And then they would need to be `sarray_remove` or something, which is confusing when all the user knows is that it's a shape. I'll sleep on this one. Maybe there are better names that make it clear they are out of place.

bd4: maybe get_inserted / get_removed, or copy_insert / copy_remove

PR #54, opened 2021-02-10
gview: better error on too many args

bd4:
This provides a better error message when doing things like
(gtensor<int, 2>).view(0, 1, _all), while still allowing
newaxis.


germasch (approved):
LGTM.

PR #58, opened 2021-02-11
WIP: Add a 0D assigner

td-mpcdf:
This one is on top of the previous one.
It adds a 0D assigner which is necessary for assigning a scalar to a ND-tensor in GENE.

### DIFF
--- a/include/gtensor/assign.h
+++ b/include/gtensor/assign.h
@@ -24,6 +24,19 @@ struct assigner
 };
 
 template <>
+struct assigner<0, space::host>
+{
+  template <typename E1, typename E2>
+  static void run(E1& lhs, const E2& rhs)
+  {
+    // printf("assigner<0, host>\n");
+    for (int i = 0; i < lhs.shape(0); i++) {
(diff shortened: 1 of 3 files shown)


bd4:
I have to think about this one... the LHS is still 1d in your implementation, and the template dimension arg is based on the LHS dimension. I think the `void assign(E1& lhs, const gscalar<T>& val)` at the bottom of `assing.h` is supposed to cover this case - you want to assign a scalar on RHS to a 1d LHS right? But the dimension equality check in the assign helper may prevent this from working. Also I think a literal is not automatically converted to gscalar, because conversion doesn't generally work with template deduction.

I think it would be helpful to define a test case for what you want to happen first, then we can figure out how best to make it work.

germasch:
I guess I'll refer @td-mpcdf to the discussion on slack between Bryce and myself, which isn't entirely resolved yet -- as with many things, it's complicated.

bd4:
We decided to implement this via a fill method, see #62

PR #59, opened 2021-02-11
tests: add blas dgemv test

bd4:


tests/test_blas.cxx
```diff
@@ -144,3 +146,77 @@ TEST(copy, zcopy)
...
+TEST(axpy, dgemv)
+{
+  constexpr int N = 16;
+  using T = double;
+  T* h_A = gt::backend::host_allocator<T>::allocate(N * N);
+  T* d_A = gt::backend::device_allocator<T>::allocate(N * N);
+  T* h_x = gt::backend::host_allocator<T>::allocate(N);
+  T* d_x = gt::backend::device_allocator<T>::allocate(N);
+  T* h_y = gt::backend::host_allocator<T>::allocate(N);
+  T* d_y = gt::backend::device_allocator<T>::allocate(N);
```
germasch, line 159: I think it'd be nicer to just use 1-d / 2-d `gtensor`s here, which makes the initialization and checking less awkward, or am I missing something. One could then pass `.data()` to the actual `dgemv` call.

A complication might be that one might have to use `.data().get()` for the device version of it, when thrust is used underneath. If so, that's a more general annoyance, where we might change things so that `data()` returns a raw pointer always.

tests/test_blas.cxx
```diff
@@ -144,3 +146,77 @@ TEST(copy, zcopy)
...
+  for (int i = 0; i < N; i++) {
+    // Since a*x is just ones, a*M*x computes row sums.
+    // Based on construction of M, row sums can be computed by
+    // calculating difference of triangle numbers - first row is triangle
+    // number (N-1), T_{N-1} = (N-1)*N/2.0. Second row is
+    // T_{2N-1} - T_{N-1}, third row is T_{3N-1} - T_{2N-1}, etc
+    row_sum = ((i + 1) * N * ((i + 1) * N - 1) - i * N * (i * N - 1)) / 2.0;
+
+    // b*y is just i, so total is (row_sum + i)
+    EXPECT_EQ(h_y[i], static_cast<T>(row_sum + i));
```
germasch, line 213: I'd say this is a case where conversion, ie., `T(row_sum + i)` is appropriate and not as ugly ;) Not actually sure whether that's even required, either, though.

germasch (changes requested):
I think it's fine as-is, but I see room for improvement ;)

tests/test_blas.cxx
bd4: I'm trying to remember why I did it this way. I think I was trying to make it not depend on gtensor, so this could be layered as a separate sub-library. E.g.. gtcore, gtensor, and gtlibs, where gtensor and gtlibs depend on core but not on each other. The complex type business is still super ugly though - gt::complex could be part of core, and used by the blas/fft layers, and then internally do casting to the (bitwise compatible arleady) native lib types.

bd4: We discussed this on slack - will update all the tests to use gtensor in a later PR. We decided that even if gpublas.h header has no dependency on gtensor, it's ok if the tests do. The idea of layering is not so separate into different repos, but to make sure it's easy to include headers with only what you need.

germasch (approved):

PR #60, opened 2021-02-11
Pr/scalar assign

bd4:


tests/test_assign.cxx
```diff
@@ -0,0 +1,185 @@
...
+TEST(assign, gtensor_6d)
+{
+  gt::gtensor<int, 6> a(gt::shape(2, 3, 4, 5, 6, 7));
+  gt::gtensor<int, 6> b(a.shape());
+
+  int* adata = a.data();
+
+  for (int i = 0; i < a.size(); i++) {
+    adata[i] = i;
+  }
```
germasch, line 16: Wouldn't `flatten` be a nicer way of doing this instead of the raw data access? And if we had an `atoi` (I don't think we do?), that could be made to look nicer yet.

germasch (changes requested):
Well, mostly I think we have to agree on the semantics of this, and I don't think we're on the same page yet, which is to say that I don't think this case should behave like this PR does.

That is, 
```cxx
gt::gtensor<float, 2> a(gt::shape(2, 3));
a.view(_all, _all) = 9001.; // should already work
a.fill(9001.); // I like this, too
a = 9001.; // I think this rightfully fails to compile
```

```cxx

bd4:
We decided on a different approach, mainly adding a fill method to all gtensor types. Since there is little overlap, closing this and creating a new branch/PR.

PR #61, opened 2021-02-15
ci: make sycl version independent

bd4:
Installs latest version, which has it's own drawbacks but the
repo does not keep older versions anyway. Note that this is not
as robust as it could be, because package names could still
change. Installing all of basekit would be more robust as the
documented way of installing, but it would slow the build down
a lot.


germasch (approved):
LGTM

PR #62, opened 2021-02-15
Pr/fill

bd4:


include/gtensor/gcontainer.h
```diff
@@ -87,6 +90,12 @@ inline D& gcontainer<D>::operator=(const expression<E>& e)
   return derived();
 }
 
+template <typename D>
+inline void gcontainer<D>::fill(const value_type v)
+{
+  assign(derived(), scalar(v));
+}
+
```
germasch, line 98: So this confused me, mostly the difference between `assign` and `operator=`, but I refreshed my mind, and I see that there is a specific `assign` for the r.h.s. being a scalar, and that it was me who put that in.

I think as an implementation detail that's kinda fine, even though a bit confusing, since `operator=` doesn't support a scalar r.h.s., as I think we agreed on, at least not for containers.

Using that particular `assign` as an implementation for `fill` is fine, though maybe it really should be called `fill`, or something like that, then.

It brings up the question, which doesn't necessarily have to get resolved here, what else relies on this overload of assign (assuming I put it in for a reason), because that may be a usage we may not want to continue to support (or maybe we do, but some decision should be made, and then it should be used consistently). 

Similarly, I see this in the other `assign` implementation
```cxx
  // FIXME, need to check for brodcasting
#if 0
  if (lhs.shape() != rhs.shape()) {
    std::cout << "not the same shape! " << lhs.shape() << rhs.shape() << "\n";
  }
  assert(lhs.shape() == rhs.shape());
#endif
```
which also brings up the question: Do we support broadcasting, or only broadcasting from a scalar, or not at all, and why is this in there? Again, not directly related to `fill`, but it's an issue that should be investigated and clarified (and I really should have opened an issue for it instead of typing it here...)

germasch (approved):

include/gtensor/gcontainer.h
bd4: Adding an optimized version that uses low level fill for the cases where it works is the next step I think, but using assign for the default, non-optimized works everywhere implementation makes sense, because there is a lot of code re-use - and it works because gscalar doesn't check it's index args and always returns the same value. We could rename the helper method to be `fill`, and that internally happens to call detail::assigner::run or whatever it is. That may be a bit cleaner.

germasch: I think it wouldn't be a bad idea to rename that `assign` to `fill`, but I suspect something may break when doing this, since that `assign` likely has some existing usage, so it should probably be done separately, and tested that it doesn't break gene.
(Which I guess brings up a bit of another issue, ie., having the option for gene to use a separately installed gtensor rather than the in-tree clone...)

bd4: I'm thinking a separate `include/gtensor/fill.h`, which has backend specific implementations (thrust / SYCL) for known contiguous memory blocks, and a generic implementation that calls over to assign. But I'd like to save that for another PR, the main thing now is to have a working mechanism, and if it becomes a performance bottleneck somewhere we can always prioritize using thrust::fill etc when possible.

PR #63, opened 2021-02-16
reductions: fix nothrust HIP and CUDA 10.2

bd4:


germasch (approved):
Looks good, but it'd be great if there was a less repetitive solution.

Something a long the line of
```cxx
using P = gt::thrust_pointer_t<const T,  typename Container::space_type>;
P begin = a.data(), end = a.data() + a.size();
// ...
```
which of course needs an implementation of `gt::thrust_pointer_t`, too. If this works, one wouldn't need things to be duplicated.

germasch:
👍 
Looks like it needs rebasing, though.

PR #64, opened 2021-02-16
Better performance for small 1D assignments and launches with CUDA

td-mpcdf:
If the linear size of a 1D assign or launch is less than the default blocksize BS_1D of 256, use a smaller blocksize to make it possible to have several blocks. This can increase performance a bit due to better latency hiding.

### DIFF
--- a/include/gtensor/assign.h
+++ b/include/gtensor/assign.h
@@ -221,7 +221,10 @@ struct assigner<1, space::device>
   static void run(E1& lhs, const E2& rhs)
   {
     // printf("assigner<1, device>\n");
-    const int BS_1D = 256;
+    int BS_1D = 256;
+    if (lhs.shape(0) < BS_1D) {
+      BS_1D = 32;
+    }
     dim3 numThreads(BS_1D);
     dim3 numBlocks((lhs.shape(0) + BS_1D - 1) / BS_1D);
 
--- a/include/gtensor/gtensor.h
+++ b/include/gtensor/gtensor.h
@@ -320,7 +320,10 @@ struct launch<1, space::device>
   template <typename F>
   static void run(const gt::shape_type<1>& shape, F&& f)
   {
-    const int BS_1D = 256;
+    int BS_1D = 256;
+    if (shape[0] < BS_1D) {
+      BS_1D = 32;
+    }
     dim3 numThreads(BS_1D);
     dim3 numBlocks((shape[0] + BS_1D - 1) / BS_1D);
 


germasch (changes requested):
Do you have a benchmark or some other motivation why this is important? Since we don't use anything like `__syncthreads()` that synchronizes inside a block, I don't quite see how multiple blocks would help, and we're talking tiny problems here in the first place.

germasch:
It'd be great if @mhrywniak had input on this ;)

include/gtensor/assign.h
```diff
@@ -221,7 +221,10 @@ struct assigner<1, space::device>
   static void run(E1& lhs, const E2& rhs)
   {
     // printf("assigner<1, device>\n");
-    const int BS_1D = 256;
+    int BS_1D = 256;
+    if (lhs.shape(0) < BS_1D) {
+      BS_1D = 32;
```
bd4, line 226: warp size on AMD gpus is 64, so I don't think this is portable

td-mpcdf: For AMD, BS_1D can be 64 for small dimension. I think there are also some other places where we differentiate between the architectures.

bd4: yes we can use #defines for this, maybe something in include/gtensor/device_backend.h, GTENSOR_MIN_BLOCK_SIZE, where there are already device specific defines. Or in macros.h, if that would create a circular header dependency.

bd4:
From a quick googling, I don't see anything about performance of small blocks. In general, it would be great if we start adding benchmarks for when we do performance optimization - at least then we know what we are optimizing for. As other apps use gtensor, there may be cases where optimizing for one app slows down another, without a further rethinking, so knowing what case we optimized for is very helpful.

bd4:
Is there a specific case in GENE that this is trying to improve? I can take a stab at writing a benchmark case for this, but wondering what the original motivating case is.

mhrywniak:
> 
> 
> It'd be great if @mhrywniak had input on this ;)

I'm not sure of the circumstances of the launching code, but it looks like this change only takes effect when you have a launch that in total has fewer than the default 256 elements (right?). So in this case instead of launching a large block with potentially idle threads, you launch multiple smaller blocks concurrently. 

Those could go to multiple SMs (instead of the large block, which will be resident on a single SM), which might explain what Tilman is seeing. I agree some (micro)benchmark/sample launch would help.

germasch:
I guess I'm closing this due to inactivity...

PR #65, opened 2021-02-18
Pr/better blas interface

bd4:


bd4:
This is a major refactor, implementing a C++ style blas interface in `gt::blas`. It includes a low level interface, mirroring the underlying blas routines and taking device pointers to arrays, and a simplified mid-level interface that takes gtensor or gtensor_span objects. Rather than implementing all the backends in one file, code is moved to `include/gtensor/blas/{backend}.h`. Currently only cuda is implemented.

Some open questions:

1. Should we move to an object oriented interface, rather than passing a handle (full C++ style). This could be for all the routines, low and mid level, or for mid level only.
2. Should we follow cublas convention and accept a pointer for the scalars (e.g. in scal and axpy), which can be either host or device, or take the parameter by value (always host). I'm inclined to the latter, because MKL does not support passing a device pointer anyway so it's not portable, and it's awkward unless one digs into the docs to understands the underlying reason.
3. Should we remove the low-level interface, and have the C interface layer use gt::adapt to call in to the mid-level interface? My inclination is not to do this, because it makes blas have a hard dependency on the gtensor part of the headers, and the it so happens that the low-level vs mid-level interfaces map nicely into backend specific vs backend agnostic layers.

include/gtensor/blas/cuda.h
```diff
@@ -0,0 +1,136 @@
...
+  {                                                                            \
+    gtGpuCheck((cudaError_t)METHOD(h, n, (BLASTYPE*)a, (BLASTYPE*)x, incx,     \
+                                   (BLASTYPE*)y, incy));                       \
+  }
+
+CREATE_AXPY(cublasZaxpy, gt::complex<double>, cuDoubleComplex)
+CREATE_AXPY(cublasCaxpy, gt::complex<float>, cuComplex)
+CREATE_AXPY(cublasDaxpy, double, double)
+CREATE_AXPY(cublasSaxpy, float, float)
+
```
germasch, line 68: I think it's good style to `#undef` the macro after you're doing with it, not that it's likely that it'll ever conflict with anything.

You could consider making a single macro that does all of axpy, scal, etc, then you have to only invoke it once, though I guess it'd need some `cublas ## TYPE ## axpy` kinda thing, but I don't mind it the way it is.

include/gtensor/blas/cuda.h
```diff
@@ -0,0 +1,136 @@
...
+template <typename T>
+inline void scal(handle_t h, int n, const T fac, T* arr, const int incx);
+
+#define CREATE_SCAL(METHOD, GTTYPE, BLASTYPE)                                  \
+  template <>                                                                  \
+  inline void scal<GTTYPE>(handle_t h, int n, const GTTYPE fac, GTTYPE* arr,   \
+                           const int incx)                                     \
+  {                                                                            \
+    gtGpuCheck(                                                                \
+      (cudaError_t)METHOD(h, n, (BLASTYPE*)&fac, (BLASTYPE*)arr, incx));       \
```
germasch, line 81: C-style casts are frowned upon by some people (e.g., hard to grep for), so `static_cast` wouldn't hurt ;)

include/gtensor/cblas.h
```diff
@@ -1,5 +1,37 @@
...
+void gtblas_caxpy(int n, const gtblas_complex_float_t* a,
+                  const gtblas_complex_float_t* x, int incx,
+                  gtblas_complex_float_t* y, int incy);
+void gtblas_zaxpy(int n, const gtblas_complex_double_t* a,
+                  const gtblas_complex_double_t* x, int incx,
+                  gtblas_complex_double_t* y, int incy);
+
+#ifdef __cplusplus
+}
+#endif
```
germasch, line 32: I'm a fan of `{BEGIN,END}_C_DECLS`, though it looks like that would have to be added to gtensor, it's currently only  in gene.

src/cgtblas.cxx
```diff
@@ -1,107 +1,49 @@
...
-    // TODO: exception handling
-    auto e = oneapi::mkl::blas::axpy(*handle, n, a, x, incx, y, incy);
-    e.wait();
-#endif
-}
-*/
+#if 0
 
-void gpublas_zdscal(int n, const double fac, gpublas_complex_double_t* arr,
+void gtblas_zdscal(int n, const double fac, gtblas_complex_double_t* arr,
```
germasch, line 46: Is there a reason for introducing `gtblas_complex_double_t` as opposed to just using `gt::complex<double>`? It's kinda unfortunate that we can't just use `std::complex` (maybe eventually), but do we need yet another wrapper?

tests/test_blas.cxx
```diff
@@ -2,11 +2,147 @@
...
+    h_x(i) = T(2.0 * i, -2.0 * i);
+  }
+
+  gt::copy(h_x, d_x);
+
+  gt::blas::handle_t h;
+
+  gt::blas::create(&h);
+
+  gt::blas::scal(h, N, a, gt::backend::raw_pointer_cast(d_x.data()), 1);
```
germasch, line 123: Not as part of this PR, and I probably said this before, but I think it's probably worth exploring hiding whether we use thrust::vector or not as the backend storage in gtensor's API, ie., return the raw pointer from `.data()` regardless. There's many more places, though, where we currently end up with `thrust::device_reference`, so this isn't a trivial undertaking, but I think it could also help avoiding some annoying problems that I kinda remember where the automatic conversion to the underlying type doesn't work.

germasch (approved):
Looks quite nice to me ;). I only have a couple of minor comments.

bd4:
Known SYCL/MKL issue: all `dot` routines fail with `"Native API failed. Native API returns: -30 (CL_INVALID_VALUE) -30 (CL_INVALID_VALUE)"`. I think this is a real MKL bug, not an issue with the calling interface. getrs also continues to fail (acknowledged bug).

PR #66, opened 2021-02-24
fix debug builds

bd4:


germasch (approved):
Seems like a good idea to add debug build to the CI ;)

I'm usually also using assert in the same fashion. However, I think it'd be better to have those checks happen unconditionally -- they are cheap compared to the actual operation, and if they trigger, they may not trigger because of a bug internal to gtensor, but rather because of someone using the API wrong, and it'd be much easier to let them know even in production mode.

bd4:
That is a good point - I guess we need to define our own gt::assert for this, for unconditional asserts not affect by NDEBUG?

germasch:
There is `Expects`, not standard yet I believe. I think this may be what we want. Or we could just throw exceptions, but it's a bit lengthy to have those all over the place.

bd4:
Looks like Microsoft GSL has a standalone header we could drop in: https://github.com/microsoft/GSL/blob/main/include/gsl/assert

germasch:
Yeah, looks like that should work (in particular if it's usable standalone, I haven't checked)

PR #67, opened 2021-03-01
Pr/sycl use ilp

bd4:


src/gpufft.cxx
```diff
@@ -8,13 +8,16 @@
 #include "gtensor/gpufft.h"
 
 #ifdef GTENSOR_DEVICE_SYCL
+// Shouldn't be necessary with MKL_ILP64 set, but for some reason it's not
+// compiling without it.
+#define MKL_LONG std::int64_t
```
germasch, line 13: Is this another bug in the MKL? Seems like it'd be preferable to not having to add another hack.

include/gtensor/blas/sycl.h
```diff
@@ -83,29 +83,49 @@ inline void copy(handle_t* h, int n, const T* x, int incx, T* y, int incy)
 template <typename T>
 inline T dot(handle_t* h, int n, const T* x, int incx, const T* y, int incy)
 {
+  sycl::queue& q = *(h->handle);
+
+  T* d_rp = sycl::malloc_device<T>(1, q);
```
germasch, line 88: Doing these repeated allocs/frees seems like a potentially noticeable performance issue. How about allocating this statically at device init time?

germasch (changes requested):
LGTM.

I guess I'm not happy with LP64 being broken. For functions that need an array of ints, that means one has to either adapt GENE to use conditionally use `integer(8)`, or somehow copy the index arrays. Fortunately, that's probably only the case with the sparse solver stuff right now.

include/gtensor/blas/sycl.h
bd4: That is a good idea, but need to think about concurrency, or just assume only one application thread. For now I just want to have the tests passing, and be able to work towards a GENE port, so not too concerned about the performance.

src/gpufft.cxx
bd4: Yes, seems to be specific to the latest version of oneAPI in the ubuntu 20.04 apt repos, I am not seeing this on JLSE. Unfortunately there will be things like this until it stabilizes.

include/gtensor/blas/sycl.h
bd4: Hmm, would just making the variable static work for this case, so it would only be allocated on first call and re-used? Seems a little ugly in that it would show up in leak checkers since it would never be deallocated, but practically speaking, it's only a handful of bytes per function.

bd4: actually that would break the API in some sense, because the queue (and device) could be different, so it's not really a global static unless we are truly in a one fixed device per MPI process mode. While that is the default, it's possible to use a different queue with `gt::blas::set_stream`. My inclination is to leave as is until we have a real use case to test this with. Perhaps the best way to address this would be to use a pool allocator layer, rather than directly calling sycl allocator.

src/gpufft.cxx
germasch: Hmmh, hopefully this will get resolved. A particular reason I don't like the hack is that it will become incorrect if one uses LP64 again. But anyway, fine for now.

include/gtensor/blas/sycl.h
germasch: Right, and I was thinking that we only have a single static queue, but that's not at this level, but rather in the GENE-specific interface.

The correct way would be to use the handle struct to store the allocated extra memory, that should avoid concurrency issues, lack of dealloc, etc. But I agree that it's not a priority at this point, and maybe the blas calls we make are so slow that the extra alloc/dealloc will never matter. (Though if cuda is any indication, those can be quite slow.)

germasch (approved):

bd4:
Latest commit uses the group API for getrf/getrs, which is more similar to the cuBLAS API, e.g. it takes as input an array of pointers to matrices in device memory. Unfortunately it is not identical - it also expects the pivot array to be an array of pointers in device memory. As an expedient way to get this working and passing tests, I allocate shared memory for this purpose on every call. With a pooled allocator backend, this might be a reasonable way of doing it. Alternatively the wrapper API could be changed to take both a device array of device pointers, and a host device pointer to the start of pivot array, which would allow all underlying APIs to work, at the cost of some unnecessary allocations in application code. This is not a good candidate for allocating as part of the gtblas handle, because it's size depends on the batch size. I guess we could impose a max batch size and pick some reasonable size array... but I think pooled allocation is better solution.

PR #70, opened 2021-03-11
Pr/fft refactor

bd4:


CMakeLists.txt
```diff
@@ -321,20 +321,29 @@ endif()
...
   message(STATUS "${PROJECT_NAME}: FFT is ENABLED")
-  add_library(gpufft)
-  target_gtensor_sources(gpufft PRIVATE src/gpufft.cxx)
-  target_link_libraries(gpufft gtensor::gtensor)
+  add_library(gtfft INTERFACE)
+  target_link_libraries(gtfft INTERFACE gtensor::gtensor)
 
   if (${GTENSOR_DEVICE} STREQUAL "cuda")
-    target_link_libraries(gpufft "cufft")
+    target_link_libraries(gtfft INTERFACE "cufft")
```
germasch, line 328: Whereas I don't think it hurts, there's no reason to put the name of the library in quotes. 

In any case, not sure we're requiring a new enough cmake yet, but there's a better way to find cufft now: https://cmake.org/cmake/help/git-stage/module/FindCUDAToolkit.html

include/gtensor/fft.h
```diff
@@ -0,0 +1,79 @@
...
+  REAL,
+  COMPLEX
+};
+
+enum class Direction
+{
+  FORWARD,
+  INVERSE
+};
+
```
germasch, line 27: Do we still need this?

include/gtensor/fft.h
```diff
@@ -0,0 +1,79 @@
...
+  template <typename C1, typename C2,
+            typename = std::enable_if_t<
+              has_container_methods_v<C1> && has_space_type_device_v<C1> &&
+              has_container_methods_v<C2> && has_space_type_device_v<C2>>>
+  void operator()(C1& in, C2& out)
+  {
+    operator()(gt::backend::raw_pointer_cast(in.data()),
+               gt::backend::raw_pointer_cast(out.data()));
+  }
```
germasch, line 63: Not sure there's an easy way to do this, but it'd be kinda nice to handle constness (both making it more explicit, and enforcing it, ie., `out` must return a pointer to a non-const type -- this may already be the case.)

include/gtensor/fft/cuda.h
```diff
@@ -0,0 +1,175 @@
...
+struct fft_config<gt::fft::Domain::COMPLEX, double>
+{
+  constexpr static cufftType type_forward = CUFFT_Z2Z;
+  constexpr static cufftType type_inverse = CUFFT_Z2Z;
+  constexpr static auto exec_fn_forward = &cufftExecZ2Z;
+  constexpr static auto exec_fn_inverse = &cufftExecZ2Z;
+  using Tin = gt::complex<double>;
+  using Tout = gt::complex<double>;
+  using Bin = cufftDoubleComplex*;
+  using Bout = cufftDoubleComplex*;
```
germasch, line 28: There may be a reason, but I guess I'd find it preferable to define the type without the `*` here, for symmetry between `Tin` and `Bin`.

include/gtensor/fft/cuda.h
```diff
@@ -0,0 +1,175 @@
...
+    auto type_forward = detail::fft_config<D, R>::type_forward;
+    auto result =
+      cufftPlanMany(&plan_, rank, n, nullptr, istride, idist, nullptr, ostride,
+                    odist, type_forward, batch_size);
+  }
+
+  virtual ~FFTPlanManyCUDA() { cufftDestroy(plan_); }
+
+  void operator()(typename detail::fft_config<D, R>::Tin* indata,
+                  typename detail::fft_config<D, R>::Tout* outdata)
```
germasch, line 144: So it does look like passing a const * for `indata` won't work, and while that may be a reflection of the original interface, it'd be kinda nice to do this correctly in our own interface.

tests/test_cfft.cxx
```diff
@@ -0,0 +1,256 @@
...
+  // but with fftw convention for real transforms, the last term is
+  // conjugate of second and set to 0 to save storage / computation
+  execf(p, gt::backend::raw_pointer_cast(d_A.data()),
+        gt::backend::raw_pointer_cast(d_B.data()));
+
+  delf(p);
+
+  gt::copy(d_B, h_B);
+
+  expect_complex_eq(h_B(0, 0), T(8, 0));
```
germasch, line 56: What's the reason again that `EXPECT_EQ()` doesn't work here?

germasch (changes requested):
LGTM, with I'd say the only thing that should be changed for this PR being the `"cufft"` lib.

CMakeLists.txt
bd4: Current min version is 3.13; FindCUDAToolkit was added in 3.17. In my experience, if a system is running ancient RHEL with an old cmake, they provide a module with 3.17 or greater. So I don't think it would be too limiting to bump the minimum version to 3.17. Lets keep that as a separate PR though.

include/gtensor/fft.h
bd4: Nope, good catch

include/gtensor/fft.h
bd4: I can const cast before passing in to cuFFT. For some reason none of the low level libraries using const input. I'll try and see if it works.

tests/test_cfft.cxx
bd4: It should really be called `expect_complex_near`. I think the issue was that `EXPECT_NEAR` did not work with complex.

bd4: renamed

include/gtensor/fft/cuda.h
bd4: added const to indata type in C++ and C interfaces

include/gtensor/fft/cuda.h
bd4: good point, removed pointer

germasch (approved):

bd4:
Design change - moved fft includes to `include/gt-fft`, to make it more clear that this is an optional add on when looking at the code (and could be separated into a different repo if we ever want/need to).

Also considering removing the CLib portion, and making that part of GENE only. Then gtensor becomes a C++ library only. The drawback is it will be a bit more work to set up testing for the CLib layer.

include/gt-fft/backend/sycl.h
```diff
@@ -0,0 +1,169 @@
...
+      plan_->set_value(oneapi::mkl::dft::config_param::CONJUGATE_EVEN_STORAGE,
+                       DFTI_COMPLEX_COMPLEX);
+      plan_->commit(gt::backend::sycl::get_queue());
+    } catch (std::exception const& e) {
+      std::cerr << "Error creating dft descriptor:" << e.what() << std::endl;
+      abort();
+    }
+  }
+
+  virtual ~FFTPlanManySYCL() { delete plan_; }
```
germasch, line 132: Since this class is probably not meant to be derived from, I think having this `virtual` may give the wrong impression.

In any case, I advocate using `std::unique_ptr` for the plan -- not only does it mean you don't need a destructor, it also helps to get other semantics correct (in particular, move ctor/assign should work, and copy ctor/assign will be deleted, which is preferable to what would currently happen if one tried).

include/gt-fft/backend/cuda.h
```diff
@@ -0,0 +1,190 @@
...
+      cufftPlanMany(&plan_inverse_, rank, n, nullptr, ostride, odist, nullptr,
+                    istride, idist, type_inverse, batch_size);
+  }
+
+  virtual ~FFTPlanManyCUDA()
+  {
+    cufftDestroy(plan_forward_);
+    cufftDestroy(plan_inverse_);
+  }
+
```
germasch, line 103: I commented on a similar issue later (sycl), though this comment may show up first. `std::unique_ptr` doesn't (easily) work here. But I think there's still the "rule of five" (IIRC), ie., if one has destructor, one should also define copy/move ctor/assign. Or, at least, delete them to prevent bad things from happening.

germasch (changes requested):

include/gt-fft/backend/cuda.h
germasch: I guess github actually does put comments in the order written, maybe ;) -- in any case at least in this case.

bd4: Seems like an FFT plan should be a move only type? I am curious now what semantics the MKL descriptor has.

bd4: Huh MKL declares a destructor, but no copy/move ctor/assign.

include/gt-fft/backend/sycl.h
bd4: It actually is derived from, to define the backend-neutral helper methods that take gtensor container objects. But perhaps there is a better way of doing that.

germasch: Ok -- well, nothing wrong about that, I guess there's a choice of inheritance vs composition, but I can see the point for inheritance here.

The `virtual` thing in the destructor probably still isn't needed, but arguably should be done anyway (besides that it's moot now due to using unique_ptr). This is about keeping a pointer to the base class, when the actual object is a derived class, so then when deleting the pointer, the "wrong" destructor gets called. That's probably not how one would use this class, but in any case, I agree that a class that is derived from should have a virtual destructor, just in case.

include/gt-fft/backend/cuda.h
germasch: That's usually a bug: https://en.cppreference.com/w/cpp/language/rule_of_three

But usually doesn't mean always, so I don't know.

Issue #7, opened 2020-05-15
Add numpy like gtensor create/initialize helpers


bd4:

bd4:
The gtensor constructor produces uninitialized data. This is great for performance in many cases, but it can lead to bugs in other cases. To make this more explicit, we can discourage the use of the constructor, and instead using helpers modeled after numpy, that take the shape as their only argument and return an appropriate gtensor object. For example:

```
auto a_zero_init = gt::zeros<double, 2>(gt::shape(10, 10));
auto a_no_init = gt::empty<double, 2>(gt::shape(10, 10));
```

Just like for the gtensor constructor, these will default to creating host vectors, and device vectors can be specified using the third template parameter, or the short cuts `empty_device` and `zeros_device`.

To support other fill values, we can also provide a `gt::full<T, N, S>(shape, value)` helper, again modeled after numpy. This constructs an empty object with the specified shape, the calls the `.fill(value)` method and returns the new object. We could easily provide a `gt::ones` as well, but `full` covers this case and others. `zeros` can also be build on `full`, but it is common enough to warrant it's own helper, whereas ones is not necessarily that much more common than other nonzero fill values.

bd4:
See #71

bd4:
Note, after some discussion and trying it out in test cases, we decided to have the _like routines care the space type over, rather than have _like_device. They are actually much more convenient to use that way, and arguably more intuitive than being mostly like except for the space.

PR #71, opened 2021-03-29
Pr/create helpers

bd4:


bd4:
To be merged after fft refactor

include/gtensor/memset.h
```diff
@@ -0,0 +1,39 @@
...
+{
+
+template <typename S>
+inline void memset(void* dst, int value, gt::size_type nbytes);
+
+template <>
+inline void memset<gt::space::host>(void* dst, int value, gt::size_type nbytes)
+{
+  std::memset(dst, value, nbytes);
+}
```
germasch, line 22: So overall I feel like memset is more C than C++, and C++ one wold use `std::fill`. Not sure that this is really something to care about, though. setting a value other than zero with memset is usually not particularly useful, since it really only works with byte-size underlying types.

germasch (approved):
LGTM with some minor comments.

include/gtensor/memset.h
bd4: This is also wrapping the device (CUDA/HIP) memset, which is very much a C interface. So option is to upgrade them all to C++ at the lowest possible wrapper layer, or do a thin wrap in the low level, and all public (non backend) style interface is proper C++ but may have to use a few of these low level C-like interfaces in the implementation. The problem with going straight to C++ wrapper interface, it implies a functionality not available on device, i.e. setting values that are larger than a byte.

germasch: Yeah, that's kinda of what I thought, and that's a good reason that I clicked on "Approve" in any case ;)

germasch:
LGTM. This should say "Fixes #7", right?

PR #72, opened 2021-04-02
little bits and pieces

germasch:
This add some `GT_INLINE` to functions that should be usable from the device (ie., which I needed to be usable on the device), and also adds `inline` to  `GT_INLINE` macro, which was missing in the CUDA case.

### DIFF
--- a/include/gtensor/gstrided.h
+++ b/include/gtensor/gstrided.h
@@ -51,7 +51,7 @@ public:
   using base_type::derived;
 
   gstrided() = default;
-  gstrided(const shape_type& shape, const strides_type& strides);
+  GT_INLINE gstrided(const shape_type& shape, const strides_type& strides);
 
   GT_INLINE int shape(int i) const;
   GT_INLINE const shape_type& shape() const;
@@ -77,8 +77,8 @@ protected:
 // gstrided implementation
 
 template <typename D>
-inline gstrided<D>::gstrided(const shape_type& shape,
-                             const strides_type& strides)
+GT_INLINE gstrided<D>::gstrided(const shape_type& shape,
+                                const strides_type& strides)
   : shape_(shape), strides_(strides)
 {}
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+build*/
+.vscode/
--- a/include/gtensor/gtensor.h
+++ b/include/gtensor/gtensor.h
@@ -324,6 +324,66 @@ struct launch<3, space::host>
   }
 };
 
+template <>
+struct launch<4, space::host>
+{
+  template <typename F>
+  static void run(const gt::shape_type<4>& shape, F&& f)
+  {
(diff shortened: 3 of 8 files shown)


germasch:
@bd4 I'm kinda at a loss as to why there are so many CI failures -- I don't think they're because of the actual changes in this PR...

include/gtensor/gstrided.h
```diff
@@ -51,7 +51,7 @@ class gstrided : public expression<D>
   using base_type::derived;
 
   gstrided() = default;
-  gstrided(const shape_type& shape, const strides_type& strides);
+  GT_INLINE gstrided(const shape_type& shape, const strides_type& strides);
```
bd4, line 54: Do we ever really want/need to have constructors execute on device?

bd4 (changes requested):
Host compiler failures might be because github flipped the switch on default ubuntu 20.04. I will work on a PR that fixed that.

The HIP failure I am more concerned about. I think just need GT_INLINE on the definition outside the class, as well as the declaration inside the class.

include/gtensor/gstrided.h
germasch: Yes, I think that's sometimes useful (I think I hit that case). For example, if one wants to take a slice and then uses that subsequently. It's of course not free, but relatively cheap to do that, even in device code.

On the other hand, I don't think we want to do anything on the device that involves dynamic memory allocation.

You're right, though that there's an issue in where to put `GT_INLINE`, and what it means. It doesn't actually make sense to "inline" something that's a declaration only -- but it does need the `__host__ __device__` there. I guess at as long as the compiler doesn't complain about the superfluous `inline` it doesn't hurt to have this all in one macro. But yeah, need to look at adding it to the definition, too.

bd4: Unfortunately the ROCM/HIP stack is more picky here. We'll see what happens after rebasing on the CI fix, but I think the HIP failure is real.

germasch: Took me more attempts than it should have, but got it right finally...

bd4 (approved):

bd4:
LGTM

germasch:
I realize it isn't good style to add stuff after the fact, but I did anyway :(

I added launch support on host for 4, 5, 6-d calculations, and also made the `gtensor_span` ctor usable from the device. Ie., this should be fairly safe, and the CI really has been good at noticing problems that I missed...

germasch:
Also, I think I added `GT_INLINE` to `sarray::operator==` etc, and I don't think that was actually working since it uses `std::equal`, which isn't defined on the device (which is kinda annoying). Primarily this means I should have added tests to make sure that stuff really works..

It did cause warnings, which is how I noticed it. I wonder whether we should try to get this warnings-clean and do the CI with `-Werror` (if that even works with nvcc)...

PR #73, opened 2021-04-02
ci: use 18.04 for older compilers

bd4:
github switched to ubuntu-latest = 20.04


germasch (approved):
I guess since we specified `ubuntu-latest`, it's okay that github switches that under us. I guess I'm still unhappy though that the 18.04 image changed, too (?) Anyway, I don't think it's too much of a loss too skip a couple of compilers.

PR #76, opened 2021-04-16
fft: remove const on input

bd4:
cuFFT, particularly starting with CUDA 11, will modify input
with out of place transform on c2r. None of the underlying
libraries advertise const input, so to be safe we will remove
this promise as well, even though it appears to be true for
the other transforms.


germasch (commented):
LGTM

germasch (approved):

PR #77, opened 2021-04-16
allow adapt() to be called on the device

germasch:
It's probably not what usually should be done, but it may sometimes make sense,
if only to work around some other limitation like I'm currently doing
(the other limitation being that one can't createa view on the device,
which probably should be enabled, too.)


bd4 (approved):
LGTM

bd4:
This causes warnings about sarray constructor that calls std::copy being called from device function. I am not sure this would even work if called from device?

PR #79, opened 2021-04-17
add some unary functions and norm_linf

germasch:
They're kinda related, `gt::abs` is needed to calculate the infinity norm, and the infinity norm is useful to check results of
`gt::sin` as an example.

### DIFF
--- a/include/gtensor/gfunction.h
+++ b/include/gtensor/gfunction.h
@@ -393,12 +393,24 @@ MAKE_BINARY_OP(divide, /)
 #undef MAKE_BINARY_OP
 
 #define MAKE_UNARY_FUNC(NAME, FUNC)                                            \
+                                                                               \
+  namespace funcs                                                              \
+  {                                                                            \
+  struct NAME                                                                  \
+  {                                                                            \
+    template <typename T>                                                      \
+    GT_INLINE auto operator()(T a) const                                       \
+    {                                                                          \
+      return FUNC(a);                                                          \
+    }                                                                          \
+  };                                                                           \
+  }                                                                            \
+                                                                               \
   template <typename E,                                                        \
             typename Enable = std::enable_if_t<has_expression<E>::value>>      \
   auto NAME(E&& e)                                                             \
   {                                                                            \
-    return function(                                                           \
-      GT_LAMBDA(auto x) { return FUNC(x); }, std::forward<E>(e));              \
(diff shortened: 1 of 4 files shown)


germasch:
There's a nice mystery in here. If you change `namespace funcs` to `namespace detail` in the unary function macro, things start failing in weird ways, and I have no idea why...

bd4 (commented):
`gt::abs` and `gt::norm` exist in complex_ops.h, to wrap thrust and std versions of those functions. Perhaps these should be moved to a different namespace? We need some way to distinguish array op functions vs low level wrappers. Complex numbers is unfortunately one of the areas where all the GPU backends can be finicky and inconsistent, i.e. sometimes std::X works and sometimes it does not in device code, sometimes it works with the right compiler options, sometimes not at all. It's getting better, but needs careful testing.

germasch:
Well, good point. I think `gt::abs` is well defined on single numbers as well as on expressions, so I think ideally we should just support both cases with overloads.

As far `gt::norm` goes, I wasn't aware of its existence, nor of what it actually does (I know now). I don't think it's used anywhere. I think there's a theoretical point for having it, since it makes it more efficient to calculate, e.g., the l2-norm. In any case, it doesn't conflict with `gt::norm_linf` -- and I'm not sure that one should be its final incarnation. xtensor has it, but it seems that the preferred way of doing norms is `xt::linalg::norm`.

So I think I should make sure that `gt::abs()` handles both native numeric types and expressions, but that should be good then?

bd4:
> So I think I should make sure that `gt::abs()` handles both native numeric types and expressions, but that should be good then?
Now that I think about it, I think `gt::abs` as defined may only work on `gt::complex`. We should probably fix that so it works on double and float too, but maybe out of scope for this PR. As long as this works with gt::complex and the gt array types, I think is ready to merge.

I will test this on AMD and SYCL as well, the CI just tests for compile and for complex numbers, sometimes issues don't show up until runtime.

bd4 (approved):
There are already tests for the overload of gt::abs and they are passing. I also tested on SYCL and HIP and the new tests are working.

germasch:
So I just added one new test that uses the new `gt::abs` on a host gtensor of type `std::complex`, and that works fine. I think that'll be good enough for the time being, while we leave the more iffy task of making this work on the device with whatever complex type for the future...

PR #87, opened 2021-04-21
try out typed tests

germasch:
This is mostly to see what the CI says, since I think this isn't ready for primetime.

I think fundamentally, typed tests are a good way to avoid duplication. My original goal was to specialize tests, for `float`, `double`, `gt::complex<float>`, etc., but that hit some early obstacles (which need more looking at, and probably should be fixed.)

So I went ahead with the other thing I was thinking to do, ie., generating tests for `gt::space::host` and `gt::space::device`. And while this works with CUDA, I think the way it is now, it assumes that device arrays can be initialized and accessed on the host, which we probably don't want to depend on in general (ie., without thrust).

But so it also is actually a good starting point for a discussion on how to have an easy interface for writing gtensor code that can be compiled into gpu or cpu code transparently.


bd4:
This looks reasonable. I did some templated tests e.g. in test_fft, but without using the googletest support - I just called a templated test method from different TEST annotated calls. This scales better to more than two tests, even if the syntax isn't the most elegant thing.

germasch:
I don't mean to say it's ready for review, I'm just wondering whether the Draft status is why the CI didn't run. Or maybe that's a consequence of the rename of the main branch....

bd4:
I guess CI config is one thing the github feature does not do yet...

PR #90, opened 2021-04-22
functions: test/support gt::abs on device

germasch:
I'm marking this WIP because it's (1) ugly and (2) uses `std::complex` on the device, which gives warnings, though it does actually succeed.

So it's kinda just for reference, if someone gets around to trying to do this right...


germasch:
I guess this confirms that while std::abs(some std::complex) worked for me, in some environments it does not, so it doesn't seem like it's an option :(

PR #91, opened 2021-04-24
use CPM as cmake package manager

germasch:
It makes it easier to deal with a choice of pre-installed packages or
downloading them on demand, including overriding things to point to
a local copy when desired for development.

It'd help more if there was more than googletest that we use it for, but
that's it at this time. At least the googletest it downloads now doesn't
have the annoying warnings about ancient cmake.


bd4 (approved):
LGTM. Tested fresh build dir on Arch Linux and Debian, worked nicely for getting gtest.

PR #92, opened 2021-04-24
implement 4d assign based on 1d block/thread index, benchmark

germasch:
Doing some semi-arbitrary mapping between 4-d index space and threads/blocks is fragile, e.g., 
I hit the case where blockIdx.z exceeds the allow limit.

Using 1-d index space for launch and then calculating back a 4-d index is something that could be detrimental to performance, due to the integer divisions, but at least in the included benchmark it essentially doesn't make a difference (it's actually slightly faster on V100 for the particular case.)

There's definitely more that could be done, including
* same for other dimensionality
* don't map every index to a new thread in the first place, but map to enough threads and then loop.
* avoid mapping back to multi-d in the first place where the underlying storage is contiguous (that probably needs bigger changes in gtensor to know when that is the case.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -239,7 +239,7 @@ add_library(gtensor ALIAS gtensor_${GTENSOR_DEVICE})
 add_library(gtensor::gtensor ALIAS gtensor_${GTENSOR_DEVICE})
 
 include(CTest)
-if (BUILD_TESTING AND IS_MAIN_PROJECT)
+if ((BUILD_TESTING AND IS_MAIN_PROJECT) OR GTENSOR_BUILD_BENCHMARKS)
   message(STATUS "${PROJECT_NAME}: build testing is ON")
 
   CPMAddPackage(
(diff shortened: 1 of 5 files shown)


bd4:
For SYCL, I implemented a generic N-d using the flatten helper:
https://github.com/wdmapp/gtensor/blob/main/include/gtensor/assign.h#L408
I think this is more elegant, and should have basically the same performance - but can verify with the new benchmark.

germasch:
> For SYCL, I implemented a generic N-d using the flatten helper:
> https://github.com/wdmapp/gtensor/blob/main/include/gtensor/assign.h#L408
> I think this is more elegant, and should have basically the same performance - but can verify with the new benchmark.

I agree that this is the ugly ad-hoc way of doing it, and generalizing this whole business would be a good step. But I guess my preference would be to merge this as-is (because it fixes a problem I'm encountering), and then do more detailed benchmarking and verification for a larger scale change later.

bd4 (approved):
LGTM, with the option to revisit in future with more general approach similar to what is used for SYCL.

PR #95, opened 2021-04-28
work towards standard allocator in gtensor_storage

germasch:
This is just part of some bigger revamp of things (which I haven't even completed).

My particular motivation was being able to support standard allocators not only in thrust but also gtensor_storage, so that whoever is using gtensor can choose their preferred pool allocator or whatever. However, it's not nearly there yet, instead it opened the opportunity to do some more not quite directly related refactoring.

In particular, I'd like to make `gtensor_storage` and `thrust::device/host_vector` more directly interchangeable, forming a separate layer from, e.g., the allocator. I'd also like to see `gtensor_storage`'s implementation to be less directly dependent on whether the underlying storage is on the host, the device, managed or whatever. For now, this PR makes things a bit more complicated by introducing an additional `ops` template arg that's designed to encode the things that are still dependent on host/device. Eventually, this should be able to be eliminated.

The other thing that this is starting is to make the code less `#ifdef` dependent -- e.g., there's no fundamental reason why we shouldn't be able to  support both thrust and gtensor_storage as the backend at the same time. This does work toward that by organizing some backend stuff into namespaces -- which actually won't be the final solution, since one cannot template by namespace, but it's somewhat of a step toward it.

In the "unification", I needed to make some code independent of the backend, so this is an opportunity to generally try out approaches to allow for writing truly "portable" code. I went with a `host_mirror` and `copy` approach, I believe pretty much what kokkos does. I'm not so sure that'll be my final preference, but it's non-invasive (it doesn't change existing behavior), and explicit.

### DIFF
--- a/benchmarks/ij_deriv.cxx
+++ b/benchmarks/ij_deriv.cxx
@@ -507,16 +507,16 @@ void test_ij_deriv(int li0, int lj0, int lbg0)
   // cleanup
   free(ref_darr);
 
-  gt::backend::host_allocator<Complex>{}.deallocate(h_arr);
-  gt::backend::device_allocator<Complex>{}.deallocate(d_arr);
+  gt::backend::host_allocator<Complex>{}.deallocate(h_arr, arr_size);
+  gt::backend::device_allocator<Complex>{}.deallocate(d_arr, arr_size);
 
-  gt::backend::host_allocator<Complex>{}.deallocate(h_darr);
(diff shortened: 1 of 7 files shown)


bd4 (approved):
LGTM. AMD with and without thrust is passing unit tests. SYCL is failing but I believe this was from the 4d test case introduced by previous PR, and not related to this.

PR #97, opened 2021-04-29
more allocator-related reorg

germasch:
It's kinda beyond actual allocator stuff, more of a some ongoing reorganization of the backend stuff. In particular, the copy and memset functions are now put into `struct ops` in the various `cuda`/`hip`/ etc namespaces.


bd4:
LGTM, other than test failure, maybe related to a missed rename in test_fft.cxx.

germasch:
Yup, I suspect I didn't have that enabled.

bd4 (approved):
LGTM

PR #98, opened 2021-04-29
use FindCUDAToolkit for cufft, cublas

germasch:
Sounds like generally a good idea, and specifically fixed linker errors on my Mac.


bd4 (approved):
Works for me on debian machine with 10.2 (spack) and 11.2 (system), and on summit. Makes me a little nervous changing things up with all the diverse environments, but we can always make adjustments if needed... and so far seems like this improves things at least on one platform, so maybe it will help elsewhere.

PR #99, opened 2021-04-30
gfunction fixes

germasch:
This fixes a bug in unary `gfunction::size()`, and adds it in the binary case.

This should fix one of the sycl build errors.


germasch:
Well,  looks like this PR fixes two of the sycl failures, but that's not the end of it... In any case, it also fixes a real bug, so I think it's worth merging...

germasch:
gi

bd4 (approved):
LGTM. The remaining sycl failure from ` gt::copy(gt::eval(b), h_b);` already existed in main branch.

PR #101, opened 2021-04-30
more allocator / backend related updates

germasch:
This starts namespaces to select the "default" backend (first called "standard", then I decided to go with "system"). That ends up being distinct from another selection, "clib", which is what the C interface is building upon. The former can be thrust, but the C interface goes lower than what thrust can provide, so it needs one of the actual low-level interfaces (cuda/hip/sycl).


bd4 (changes requested):
ij_deriv benchmark is failing to compile for me, with both CUDA and HIP. I guess the CI doesn't enable building it, that should be fixed as well (but can be a different PR).

germasch:
> ij_deriv benchmark is failing to compile for me, with both CUDA and HIP.

Good catch, I clearly did have benchmark build turned on at some point, but apparently not later on. Should now be fixed (with ugly `raw_pointer_cast`s, but hopefully they'll see limited use eventually...).

bd4 (approved):
Working for me now on CUDA and HIP, both tests and benchmark

PR #103, opened 2021-05-01
allocator: allow for overriding default allocator

germasch:
This separates out the caching allocator into a separate file, and introduces `GTENSOR_DEFAULT_{HOST,DEVICE}_ALLOCATOR` macros which can be set to override the default allocator.

Though by now I think that, for my purposes, I'll rather make my own variant of `gt::gtensor` that allow me to use my choice of allocator without relying on macro hackery.

### DIFF
--- a/include/gtensor/allocator.h
+++ b/include/gtensor/allocator.h
@@ -1,0 +1,122 @@
+
+#include <iostream>
+#include <map>
+#include <memory>
+
+namespace gt
+{
+namespace allocator
+{
+
+// ======================================================================
+// caching_allocator
+
+template <class T, class A>
+struct caching_allocator : A
+{
+  using base_type = A;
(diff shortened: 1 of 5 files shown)


bd4:
Seeing some strange behavior on one of my machines but not the other. Both HIP and CUDA on my Argonne machine are hanging at gtensor.device_move_ctor. CUDA on my laptop works fine with nothrust.

bd4:
Ok the good news is now I can reproduce on my laptop too (bug in my cmake script), bad news is that there is a real deadlock in nothrust backend that is triggered by the gtensor.device_move_ctor unittest.

bd4 (approved):
LGTM, once we resolve the cuda 10.0 issue (old thrust) found by the CI.

PR #104, opened 2021-05-02
finish (hopefully) allocator related work

germasch:
There is certainly some more refactoring / generalizing backends to come, but this I think is it in terms of adding flexibility to choose an allocator (as well as backend storage).

The first part is a bit of work towards having different tags for `cuda`, `hip`, etc, and having `device` just be an alias for whatever is default. However, this didn't go very far as other parts aren't really ready for it -- in particular, `gtensor_storage` is still using plain pointers, which are obviously indistinguishable between `cuda`, `hip`, etc, or even `host`.

The second part essentially turns `gtensor` into `gtensor_container`, which is now templated by the underlying storage (`EC`, take from xtensor, I suppose short for `element container`).  `gtensor` itself became an alias, preserving its template arguments, though as can be seen in the examples, that did cause a bit of breakage since template argument deduction is impacted.

### DIFF
--- a/examples/src/daxpy.cxx
+++ b/examples/src/daxpy.cxx
@@ -36,18 +36,16 @@ using namespace gt::placeholders;
  * all the data which would happen by default (just like std::vector, gtensor
  * has copy semantics by default).
  */
-template <typename S>
-gt::gtensor<double, 1, S> daxpy(double a, const gt::gtensor<double, 1, S>& x,
-                                const gt::gtensor<double, 1, S>& y)
+template <typename Ex, typename Ey>
+auto daxpy(double a, const Ex& x, const Ey& y)
 {
   // The expression 'a * x + y' generates a gfunction template, which is an
-  // un-evaluated expression. In this case, the return statement and return
-  // type force evaluation so a new gtensor is created. It can be useful to
-  // instead set the return type to 'auto', in which case the un-evaluated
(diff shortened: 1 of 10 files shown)


bd4:
HIP thrust/nothrust are working. CUDA nothrust, with gcc 8, is failing compile with the following error:
```
/home/bda/fusion/gtensor/tests/test_gtensor.cxx:238:70: error: ‘template<class _Tp> using __pointer = typename _Tp::pointer’ is protected within this context
   GT_DEBUG_TYPE_NAME(decltype(h1)::pointer);
                                                                      ^                                                                                                                                                                    
/usr/include/c++/8/bits/alloc_traits.h:56:61: note: declared protected here
     template<typename _Tp>
```

bd4:
Seems to come from the fact that pointer is now based on allocator traits, via gtensor_storage:
```
using pointer = typename std::allocator_traits<A>::pointer;
```
Seems like this should work, making a trait protected seems broken by design. So not sure if this is a bug in the stdlib implementation for this particular gcc or what.

bd4 (approved):
LGTM. Seeing a compile issue on one x86_64 with older cuda toolkit and older gcc. However can't reproduce anywhere else, and it's in a debug macro, so I think we can safely merge this and revisit in future if it comes up elsewhere.

PR #105, opened 2021-05-05
introduce gt::device_ptr

germasch:
Even for the case without thrust, it's quite useful to be able to distinguish which memory space a given pointer points to. The goal is not to allow for (kinda) transparent access to device memory from the host, like thrust, but essentially just to have tagged pointers. This will allow having to keep track of the memory space separately, as is currently the case, where if one calls, e.g. `copy`, one needs to provide the spaces in addition to the pointers.

Only `device_ptr` itself is a fancy pointer type -- when dereferencing it, one ends up with a regular `T&` reference. That's good in that everything then works as usual without any trickery, though one needs to be aware that `&*ptr` hence loses the memory space information and so does the same as `raw_pointer_cast`. It'd be possible to add a `device_reference` type if one wants to avoid that, but one than has to go through the troubles of making sure that that type behaves like a regular reference, which isn't completely achievable, and may not be worth it.


bd4:
Getting a segfault in cuda thrust `test_fft`. Working on testing other combinations.

bd4:
For cuda nothrust, fft passes, but test_gtensor fails `device_init_like_helpers` and `device_init_helpers_literal_shape`

bd4:
hip thrust also segfaults in fft, and in clib (memset) and gtensor (device_fill_ctors). hip nothrust fares better actually, gtensor tests `device_fill_ctors` and `device_init_*helpers*` only are failing.

bd4 (approved):
LGTM, just missing sycl->sycl copy_n to pass CI

PR #107, opened 2021-05-08
backend

germasch:
Mostly some more cleanup, moving some stuff out of `backend::`, like `gt::fill` or `gt::raw_pointer_cast`, since these functions as such are now backend-independent (and then by means of the tagged pointer are dispatched to an actual backend).

This revamps the `gallocator` abstraction a bit by separately providing `gallocator` per space type, e.g., for `cuda`, `cuda_managed` and `cuda_host` rather than having them all three nested inside of a `cuda` namespace/struct.

`gallocator` unfortunately is another abstraction level that'd be nice to get rid of -- it's used to build both the C/Fortran API as well as to make regular C++ allocators. It would have been nicer to just make C++ allocators and use them for the C/Fortran API, but that's not really possible since `deallocate` would have to know the number of bytes allocated.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -41,7 +41,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        cxx: [ 'g++-9', 'clang++-9' ]
+        cxx: [ 'g++-11', 'clang++-11' ]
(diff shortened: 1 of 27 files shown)


bd4 (approved):
LGTM, all tests passing on real GPUs for CUDA, HIP (thrust and nothrust), SYCL.

PR #108, opened 2021-05-08
build: modernize pulling in of GoogleTest

germasch:
Standard cmake calls it "GTest", and the targets also follow
a new convention.

In addition, one should create all targets, even the ones we don't need,
since this may be used in a dependent package that also wants gtest but
needs those additional targets.


bd4 (approved):
LGTM

PR #109, opened 2021-05-10
ci: use custom image for SYCL

bd4:


.github/workflows/sycl.yml
```diff
@@ -23,17 +23,11 @@ jobs:
...
-        echo "deb https://apt.repos.intel.com/oneapi all main" > /etc/apt/sources.list.d/oneAPI.list
-        echo 'deb [trusted=yes arch=amd64] https://repositories.intel.com/graphics/ubuntu bionic-devel main' > /etc/apt/sources.list.d/intel-graphics.list
-        apt-get update -y
-    - name: install dpcpp
-      run: apt-get install -y intel-oneapi-compiler-dpcpp-cpp intel-oneapi-mkl intel-oneapi-mkl-devel clinfo intel-opencl-icd intel-level-zero-gpu level-zero
     - uses: actions/checkout@v2
+    - name: install packages
+      run : |
+        apt-get update -y
+        apt-get install -y --no-install-recommends clang-9
```
germasch, line 30: Why don't you install clang-9 into the custom image as well?

bd4: Good point, not too big I think even when it pulls in a fair number of packages. I updated the image and force pushed removing this.

PR #117, opened 2021-05-22
add fused vs non fused benchmark

germasch:
This models some of the operations used by GENE.

Against my better judgement, this only adds the benchmarks, without actually testing that the results are correct ;(


bd4 (approved):
LGTM. Note that google benchmark is only working for me with CUDA. Even with the original assign benchmark, on HIP it hangs in a sched_yield() loop, and on SYCL, it fails to compile. Not needed for this PR, but at some point would be good to have it working on the other platforms.

PR #118, opened 2021-05-26
bench: add bench_hypz

germasch:
This is a sample kernel taking from GENE that has a larger stencil, calculating some hyperviscosity term.

This is actually entirely untested right now (well, it does compile with CUDA, at least). I took most of the actual code from GENE, so that should be fine, but it's still possible that I screwed up the fields passed into it...


bd4 (approved):
LGTM

germasch:
Tried running it on Summit -- it doesn't crash, so I'm calling it good.

PR #119, opened 2021-06-02
build: use "option" instead of setting a cache variable

germasch:
This avoids hassles when including gtensor via `add_subdirectory()` / CPM.


bd4:
Looks like this needs a rebase? I thought the c++20 failures were already fixed. Not sure what is going on with the SYCL failure.

germasch:
I had intentionally rebased it back onto an old version of gtensor to make it actually work in GENE temporarily (for testing my new way of pulling in gtensor), but the right fix of course is to make GENE work with current gtensor `main`. Now rebased back on top of `main`.

bd4 (approved):
LGTM

PR #120, opened 2021-06-03
don't enable GT_BOUNDSCHECK based on NDEBUG

germasch:
This bounds checking is expensive, and one might want to build one's code
with asserts enabled, but without incurring the bounds checking overhead.


bd4 (approved):
LGTM, cxx20 CI failure unrelated

PR #121, opened 2021-06-08
cmake: use TARGET_DIRECTORY for source helper

bd4:
Fixes issue with CMakeLists.txt in multi-level add_subdirectory,
where target is defined in a different file.


germasch (approved):
LGTM.

PR #122, opened 2021-06-10
Pr/fix cuda warnings

bd4:


include/gtensor/gfunction.h
```diff
@@ -110,8 +110,10 @@ inline void broadcast_shape(S& to, const S2& from)
     } else if (from[end_from - i] == 1) {
       // broadcasting, from, nothing to do
     } else {
+#ifndef __CUDA_ARCH__
       throw std::runtime_error("cannot broadcast to = " + to_string(to) +
                                " from = " + to_string(from) + "\n");
+#endif
```
germasch, line 116: It might be worthwhile to do an `assert(false)` here on the device -- it'll be compiled out with `-DNDEBUG`, but it might save someone a lot of time when chasing down some odd bug. (Or it might never happen, usually this kind of stuff should run on the host, but who knows...)

germasch (commented):
LGTM.

include/gtensor/gfunction.h
bd4: Good call, I added it.

germasch (approved):

Issue #123, opened 2021-06-28
benchmarks fail to build with SYCL backend


bd4:
Two issues - need a native SYCL implementation of ij_deriv, and need to fix N-d assign when RHS is a scalar.

The issue is that the N-d implementation uses flatten, which calls `reshape(std::forward<E>(e), shape(e.size()));`, and the `e` here is `gscalar` which has no `size` method.

PR #124, opened 2021-06-28
Pr/sycl bench

bd4:


include/gtensor/gview.h
```diff
@@ -456,10 +456,35 @@ inline auto reshape(E&& _e, gt::shape_type<N> shape)
...
+};
+
+template <typename E>
+struct flatten_impl<E, false>
+{
+  static inline auto run(E&& e) { return e; }
+};
+
+} // end namespace detail
+
```
germasch, line 482: So it's not possible to call reshape on a 1-d array? I think the original problem said gscalar, is it just for gscalar that flatten didn't work?

I generally like to think of scalars as 0-d arrays, that avoids a bunch of special cases. Does `expr_dimension` return 1 for a gscalar? If it does return 0, I think it should work to make the test below `> 0`. Though even in that case, I guess I'd prefer for reshape to work, rather than having to special case flatten.

germasch:
Fixes #123, I guess (?).

include/gtensor/gview.h
bd4: It should work on a 1-d array, but it basically adds a pointless layer that does nothing, so seems better to make it a no-op. The issue here is that flatten is used by N-d assign in the SYCL backend, whereas I don't think anyone has tried or is likely to call reshape on scalar. I think make this public, rather than private to SYCL backend, is useful, because the idea of flatten is to allow not worrying about multi-d indexing when doing simple consecutive index operations.

bd4:
> Fixes #123, I guess (?).

Yes that is the immediate purpose. Hoping to get the benchmark included in Aurora test suite somehow.

include/gtensor/gview.h
bd4: I initially tried to modify gscalar to make it behave, but it became a messy rabbit hole rather quickly... so this seemed like a more expedient way of solving the SYCL issue.

germasch: > It should work on a 1-d array, but it basically adds a pointless layer that does nothing, so seems better to make it a no-op. The issue here is that flatten is used by N-d assign in the SYCL backend, whereas I don't think anyone has tried or is likely to call reshape on scalar. I think make this public, rather than private to SYCL backend, is useful, because the idea of flatten is to allow not worrying about multi-d indexing when doing simple consecutive index operations.

I'd say the pointless do nothing issue could then be handled at the reshape layer, where it could just return the 1-d array rather than another wrapper layer around it. In any case, as I think I put into some issue as a future to-do item, flatten and related things assume that the underlying structure is contiguous, which isn't the case in general. So there's going to be some real work there at some point in the future, and until then I don't mind some slightly bandaid-ish solution ;)

germasch (approved):

PR #125, opened 2021-07-02
cgtblas: handle alignment issues

bd4:
Fortran and C (and most C++) complex types are aligned on the
component types, but thrust::complex (used when GTENSOR_USE_THRUST
cmake var is set) is aligned to the pair. This can result in
segfault if a half aligned complex is passed to a function expecting
fully aligned.


src/cgtblas.cxx
```diff
@@ -1,8 +1,79 @@
...
+}
+
+// NB: convert to the gt type component-wise in case the alignment
+// requirements are different, which will be the case when the thrust
+// storage backend is used.
+template <typename T>
+inline gt::complex<T> fc2cpp_deref(const f2c_complex<T>* c)
+{
+  return gt::complex<T>(c->real(), c->imag());
+}
```
germasch, line 31: I guess I don't like the name here very much -- those pointers can be directly dereferenced in any case, this is really more about converting to the corresponding gt type, and also about dereferencing. I don't really have a better suggestion, though...

src/cgtblas.cxx
```diff
@@ -1,8 +1,79 @@
...
+  assert_aligned(f[0], sizeof(T));
+  return f;
+}
+
+template <typename T>
+inline auto cast_aligned(f2c_complex<T>* c)
+{
+  assert_aligned(c, 2 * sizeof(T));
+  return reinterpret_cast<gt::complex<T>*>(c);
+}
```
germasch, line 52: Again, I think those are more about casting to the gt (pointer) type, which happens to be have more alignment requirement (maybe, the non-complex types definitely don't).

src/cgtblas.cxx
```diff
@@ -1,8 +1,79 @@
...
+
+template <typename T>
+inline auto cast_aligned(f2c_complex<T>** c)
+{
+  assert_aligned(c[0], 2 * sizeof(T));
```
germasch, line 57: I think this should assert that they're probably aligned for what `gt::complex` requires, which may or not be `2 * sizeof(T)`.

src/cgtblas.cxx
```diff
@@ -1,8 +1,79 @@
...
+  assert_aligned(c, 2 * sizeof(T));
+  return reinterpret_cast<const gt::complex<T>*>(c);
+}
+
+template <typename T>
+inline auto cast_aligned(const f2c_complex<T>** c)
+{
+  assert_aligned(c[0], 2 * sizeof(T));
+  return reinterpret_cast<const gt::complex<T>**>(c);
+}
```
germasch, line 73: Strictly speaking, this isn't really enough, as those are usually arrays for pointers, and just checking the first for alignment doesn't mean the others are. But of course the pointer itself doesn't even tell you how many elements you'd have to check...

germasch (approved):
I think the asserts for alignment should be changed to be based on `alignof(gt::complex<...>)`.

Other than that, I think it's fine, as it fixes real problems. I don't really like the additional complexity it introduces, but I also don't really have a great idea on how to do it better. One could go fancy C++ all the way and introduce `f2c_ptr_t`, and that could just have a corresponding `operator*` that derefs to a `gt::complex` type, and it could also have a conversion operator that casts to the `gt::complex *` after alignment check.

That still wouldn't really handle the array of pointer case, still....

src/cgtblas.cxx
bd4: my thinking is to enforce the stricter alignment always, so we can find alignment issues regardless of backend being used. Allowing less aligned for certain backends doesn't seem useful, since it's not as portable

germasch: I guess I don't necessarily agree (other backends could have an allocator that does less-aligned allocations, which would be fine as long as the operations on that backend accept those less-aligned pointers. In any case, in practice none of these theoretical things will happen, so I'm fine with this as-is.

bd4: Ok, I'll merge as is for now, so we can get the fix in to GENE. I am thinking of moving the C layer out of gtensor anyway, in which case this would no longer be a public interface, so I am less inclined to iterate on the design.

PR #134, opened 2021-09-20
add an interface with strides and dists for FFTPlanManyCUDA

td-mpcdf:


germasch (approved):
LGTM

bd4:
Looks great thank you! Because the hipfft backend is basically just #defines on top of the cuda backend, this only leaves SYCL to implement. I will merge as is, and create a new PR to add some tests and SYCL support.

PR #137, opened 2021-10-04
Pr/sycl 2021 04

bd4:
Note that this is not backwards compatible with older versions of oneapi, because of the reduction API namespace change.


germasch (approved):
LGTM. (The CI failure seems like some random-something...)

PR #142, opened 2021-10-28
sycl gt-blas improvements

bd4:
Uses gtensor RAII classes for temp memory, so things are cleaned up in case of failures, and so the memory pool can be used. Also adds a limited getrf_npvt API, which has a limitation that it requires contiguous matrices in input.


bd4:
I added a contiguous check so this won't fail silently if used not as intended.

germasch (approved):
The managed_storage stuff definitely looks like an improvement over manually allocating / freeing things for sycl.

Issue #138, opened 2021-10-12
Using streams with gtensor


td-mpcdf:
It seems to be necessary to somehow make it possible to use the (CUDA) streams with gtensor. In the launch of an explicit  kernel this could be another argument, in the assignment kernels, we should add a function set_stream(streamID) and launch the assignment kernel in this stream. A set_default_stream should set it back.

germasch:
So what is the use case that you're actually thinking about? Passing a stream to `launch` wouldn't be too hard, but it'll also require wrappers for creating / destroying streams, etc -- in a way that's portable to HIP (easy) and Intel, too. And there's the rather complicated sync semantics with respect to other streams and the default stream to keep in mind. A `set_stream` function that essentially sets a default stream for subsequent gtensor kernel invocations also needs some more thinking to ensure thread safety if gtensor is used from within a (host-)threaded environment.

bd4:
FWIW I think I can do SYCL stream emulation, by creating extra queues on demand as new streams are set in the already existing global singleton class, which is currenlty used to emulate library global default device.

I don't know if HIP has a per thread default stream option? Also SYCL definitely does not, so that approach is not portable (although perhaps DPCPP has something with there compatibility tool for translating CUDA?).

I think an argument can be made that what SYCL does, with no global state and having queue objects to encapsulate a device/stream combo, is a better API - if an application wants global state, it can define a global queue. While defining it in the library is convenient, it lacks flexibility and leads to challenges when you have multiple host threads. Currently gtensor has already inherited the global state stuff from CUDA/HIP though, so I'm not sure we want to change - it basically involves passing an extra object (the queue) to every API call, or making it a method on the queue object. I guess I'm not sure if CUDA/HIP support overriding the global device though.... something to think about.

bd4:
Looks like it uses OpenCL to support other accelerators.

bd4:
@carlosenlopez do you know if a global (per process) set_stream would work with the StarPU integration? As an alternative, we could define an assign(lhs, rhs, stream=null) function, so you could just pass the stream from starpu, but loose the nice lhs = rhs syntax.

carlosenlopez:
Here it is the answer of one of the StarPU developers:

> By default, 1 host thread per GPU device is used by the CUDA driver. If 
STARPU_NWORKER_PER_CUDA > 1, this environment variable indicates how many 
host threads are used by the CUDA driver for one GPU device.
>
[...]

PR #145, opened 2021-11-05
Pr/streams

bd4:
Simple (not space templated) stream implementation for all backends, assign and launch. Does not provide a stream arg for async copy yet. Has a couple of unit tests in tests/test_streams.cxx.

Note that the SYCL approach to streams, queue objects, is arguably a better fit for a C++ library. However the main goal with the gtensor streams implementation is to allow interoperability with other libraries that already use stream directly, e.g. StarPU.

### DIFF
--- a/include/gtensor/backend_host.h
+++ b/include/gtensor/backend_host.h
@@ -60,34 +60,27 @@ inline void fill(gt::space::host tag, Ptr first, Ptr last, const T& value)
 class stream_view
 {
 public:
-  stream_view() : stream_(nullptr) {}
-  stream_view(void* s) : stream_(s) {}
+  stream_view() {}
 
-  auto get_backend_stream() { return stream_; }
+  auto get_backend_stream() { return nullptr; }
 
-  bool is_default() { return stream_ == nullptr; }
+  bool is_default() { return true; }
 
   void synchronize() {}
-
-private:
-  void* stream_;
 };
 
-class stream_wrapper
+class stream
 {
 public:
-  stream_wrapper() : stream_(nullptr) {}
+  stream() {}
 
-  auto get_backend_stream() { return stream_; }
+  auto get_backend_stream() { return nullptr; }
 
-  bool is_default() { return stream_ == nullptr; }
+  bool is_default() { return true; }
 
-  auto get_view() { return stream_view(this->stream_); }
+  auto get_view() { return stream_view(); }
 
   void synchronize() {}
-
-private:
-  void* stream_;
 };
 
 #endif
--- a/include/gtensor/backend_host.h
+++ b/include/gtensor/backend_host.h
@@ -53,6 +53,45 @@ inline void fill(gt::space::host tag, Ptr first, Ptr last, const T& value)
 } // namespace fill_impl
 
 } // namespace backend
+
+#ifndef GTENSOR_HAVE_DEVICE
+
+// streams are no-op on host (could use threads in the future)
+class stream_view
+{
+public:
+  stream_view() : stream_(nullptr) {}
+  stream_view(void* s) : stream_(s) {}
+
+  auto get_backend_stream() { return stream_; }
+
+  bool is_default() { return stream_ == nullptr; }
+
+  void synchronize() {}
+
+private:
+  void* stream_;
+};
+
+class stream_wrapper
+{
+public:
+  stream_wrapper() : stream_(nullptr) {}
+
+  auto get_backend_stream() { return stream_; }
+
+  bool is_default() { return stream_ == nullptr; }
+
+  auto get_view() { return stream_view(this->stream_); }
+
+  void synchronize() {}
+
+private:
+  void* stream_;
+};
+
+#endif
+
 } // namespace gt
 
 #endif // GTENSOR_BACKEND_HOST_H
--- a/include/gtensor/backend_sycl.h
+++ b/include/gtensor/backend_sycl.h
@@ -173,6 +173,16 @@ public:
     return queue_map_[device_id];
   }
 
+  cl::sycl::queue new_queue(int device_id)
+  {
+    if (device_id >= devices_.size()) {
+      throw std::runtime_error("No such device");
+    }
+    return cl::sycl::queue{devices_[device_id], get_exception_handler()};
+  }
+
+  cl::sycl::queue new_queue() { return new_queue(current_device_id_); }
+
   int get_device_count() { return devices_.size(); }
 
   void set_device_id(int device_id) { current_device_id_ = device_id; }
@@ -224,6 +234,13 @@ inline SyclQueues& get_sycl_queues_instance()
 inline cl::sycl::queue& get_queue()
 {
   return detail::get_sycl_queues_instance().get_queue();
+}
+
+/*! Get a new queue different from the default, for use like alternate streams.
+ */
+inline cl::sycl::queue new_queue()
+{
+  return detail::get_sycl_queues_instance().new_queue();
 }
 
 inline void device_synchronize()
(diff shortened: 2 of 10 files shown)


bd4:
Fixes #138

include/gtensor/backend_host.h
```diff
@@ -53,6 +53,45 @@ inline void fill(gt::space::host tag, Ptr first, Ptr last, const T& value)
...
+
+  auto get_backend_stream() { return stream_; }
+
+  bool is_default() { return stream_ == nullptr; }
+
+  void synchronize() {}
+
+private:
+  void* stream_;
+};
```
germasch, line 74: If no actual streams exist, do we still need `stream_`? Can't this just be an empty class, which always `is_default()`? And I'm not sure whether it needs `get_backend_stream()`, since I'd think that never gets called, since it doesn't return anything usable. So it may not need `get_backend_stream()`, but even if does for some reason, it could just return `nullptr`.

include/gtensor/backend_host.h
```diff
@@ -53,6 +53,45 @@ inline void fill(gt::space::host tag, Ptr first, Ptr last, const T& value)
...
+
+  auto get_view() { return stream_view(this->stream_); }
+
+  void synchronize() {}
+
+private:
+  void* stream_;
+};
+
+#endif
```
germasch, line 93: Is this ever actually used? In any case, as above, I don't think it really needs the `stream_` member.

include/gtensor/backend_sycl.h
```diff
@@ -173,6 +173,16 @@ class SyclQueues
     return queue_map_[device_id];
   }
 
+  cl::sycl::queue new_queue(int device_id)
+  {
+    if (device_id >= devices_.size()) {
+      throw std::runtime_error("No such device");
+    }
```
germasch, line 180: If you're checking this, you might as well check whether it's negative, too.

include/gtensor/backend_sycl.h
```diff
@@ -425,6 +442,40 @@ inline void fill(gt::space::sycl tag, Ptr first, Ptr last, const T& value)
...
+
+  auto get_backend_stream() { return stream_; }
+
+  bool is_default() { return stream_ == gt::backend::sycl::get_queue(); }
+
+  void synchronize() { stream_.wait(); }
+
+private:
+  cl::sycl::queue stream_;
+};
```
germasch, line 460: I don't know the semantics of sycl queues (are they reference counted somehow?). A view generally shouldn't make a copy, just keep a reference, but maybe the copy here doesn't really have copy semantics?

include/gtensor/gtensor.h
```diff
@@ -598,19 +598,35 @@ struct launch<N, space::device>
...
-  detail::launch<N, space::device>::run(shape, std::forward<F>(f));
+  detail::launch<N, space::device>::run(shape, std::forward<F>(f),
+                                        gt::stream_view{});
+}
+
+template <int N, typename S, typename F>
+inline void launch(const gt::shape_type<N>& shape, F&& f,
+                   gt::stream_view stream)
+{
+  detail::launch<N, S>::run(shape, std::forward<F>(f), stream);
```
germasch, line 623: Isn't it possible to avoid the duplication here by make the stream_view argument a default argument?

tests/test_stream.cxx
```diff
@@ -0,0 +1,51 @@
...
+  gt::stream_wrapper stream;
+
+  int* adata = a.data();
+
+  for (int i = 0; i < a.size(); i++) {
+    adata[i] = i;
+  }
+
+  EXPECT_NE(a, b);
+  gt::assign(b, a, stream.get_view());
```
germasch, line 21: I'd think it should be possible to have a conversion operator that gives you a view automatically. On a side note, I don't really like `gt::stream_wrapper`. Can't we just call it `gt::stream`?

germasch (changes requested):
The main thing I think will need to be fixed eventually is that, as far as I can tell, the `stream_wrapper` and `stream_view` types are fixed at compile time by means of ifdefs, but I think that can be improved later.

include/gtensor/backend_host.h
bd4: I think it just needs the default and void* ctor, will try it.

include/gtensor/backend_host.h
bd4: Allows the tests to run on host, which exercises some bits of code that wouldn't otherwise be. But I don't have a strong opinion about it - it is an odd dummy class. I could ifdef GTENSOR_HAVE_DEVICE for the tests instead.

include/gtensor/backend_sycl.h
bd4: SYCL objects like queues  have reference semantics - copies behave the same as the original and it keeps a reference count. I believe this makes it more idiomatic to just use copies, rather than adding another layer of reference semantics on an already reference semantic construction. But I'm not sure, still wrapping my head around it. I did use reference in my earlier SYCL backend code before I understood this.

include/gtensor/gtensor.h
bd4: I spent several hours trying to get this to work and didn't succeed - I don't think it's worth the complexity. I also tried with a simple example in compiler explorer and ran into similar problems - default args only work nicely with templated types in the simplest case.

tests/test_stream.cxx
bd4: Sure, `gt::stream` does sound better.

include/gtensor/backend_host.h
germasch: Well, I'm not particularly opposed to having it, but since it's not actually doing anything, I don't think it needs to have `stream_` member which looks like it could have actually meaningful values, if it really doesn't, so this is mostly about keeping these two as simple as possible.

[I'm on the other hand not entirely convinced that it's a good idea to pretend that we can create streams on the host if we really don't, but I don't have a particularly strong opinion on that right now].

include/gtensor/backend_host.h
germasch: Does it really need the `void *` ctor if it always gets passed nullptr? (Or does it not?)

include/gtensor/backend_sycl.h
germasch: That's fine -- if it already is ref counted, why not take advantage of it.

include/gtensor/gtensor.h
germasch: But currently the `stream_view` isn't actually a template, or is it?

bd4: True, it will likely work now that I dropped the template arg. But is it worth changing now to use default args when we likely want to templatize in the next revision? I want to have Carlos actually use this first and than can iterate.

bd4:
I think this is ready now. I can be persuaded to change to use default args if you feel strongly about it, but as I said this will likely need to change again if it becomes templated again.

germasch:
The update looks good. I don't think there should be a problem with default args even when using templates (see slack), so I'd prefer to not have the duplication, but it can also be fixed later. It does look like you need to rebase the whole thing on top of latest `main`, though.

include/gtensor/gtensor.h
```diff
@@ -596,39 +596,26 @@ struct launch<N, space::device>
 } // namespace detail
 
 template <int N, typename F>
-inline void launch_host(const gt::shape_type<N>& shape, F&& f)
+inline void launch_host(const gt::shape_type<N>& shape, F&& f,
+                        gt::stream_view stream = gt::stream_view{})
```
germasch, line 600: Not worth changing, but fwiw:

```
gt::stream_view stream = {}
```

should work as well and removes some redundancy.

germasch (approved):

bd4:
rebased

PR #146, opened 2021-11-18
Pr/bench getrs

bd4:


benchmarks/getrs.cxx
```diff
@@ -0,0 +1,175 @@
...
+  CT err = CT(0.0, 0.0);
+  for (int b = 0; b < batch_size; b++) {
+    for (int i = 0; i < n; i++) {
+      for (int j = 0; j < nrhs; j++) {
+        err = h_Bdata(i, j, b) - CT(i / (j + 1) * b, i * j / (b + 1));
+        if (gt::abs(err) > 0.0) {
+          std::cout << "err of " << err << " at [" << b << ", " << i << ", "
+                    << j << "]" << std::endl;
+          ok = false;
+          break;
```
germasch, line 148: Of course it's generally frowned upon, but there are situations where `goto` is actually useful ;)

germasch (approved):
LGTM. Not really a fan of what the reversed ranges look like, but I guess if that's what needs to be done, it's gotta be done...

benchmarks/ij_deriv.cxx
```diff
@@ -220,6 +222,36 @@ void ij_deriv_gt(const gt::gtensor_span<const gt::complex<Real>, 3, Space>& arr,
...
+  auto k_darr = darr.to_kernel();
+  auto k_coeff = coeff.to_kernel();
+  auto k_ikj = ikj.to_kernel();
+  gt::launch<3>(
+    arr_shape_nohalo, GT_LAMBDA(int i, int j, int klmn) {
+      k_darr(i, j, 0, klmn) = k_coeff(0) * k_arr(i + 0, j, klmn) +
+                              k_coeff(1) * k_arr(i + 1, j, klmn) +
+                              k_coeff(2) * k_arr(i + 2, j, klmn) +
+                              k_coeff(3) * k_arr(i + 3, j, klmn) +
+                              k_coeff(4) * k_arr(i + 4, j, klmn);
```
bd4, line 250: This is not producing the correct result for x derivative, and I can't figure out why.

germasch: It looks to me like the coefficient array is on the host, and the "to_kernel()" won't change that. I think I've handled this before by having the coefficient array deliberately passed by value to the kernel, ie., passing it as a parameter will make a copy that's then used on the device.

bd4: :face_palm: yes that is the issue. Trying to decide on the most elegant why to handle it - I can copy the gtensor span to a local array variable and allow the lambda to capture it by value. Ideally the compiler is smart enough to put this in constant memory, but I'm not sure it makes much difference. Or I can just not use gtensor object for coeff in the first place, and pass them around by value as a 5 element array everywhere. Leaning towards the latter, to make the intent more clear, rather than the subtle hard coded gt::space::host that is easy to miss in the arg list.

bd4: It could also be a static global variable for that matter.

germasch: I don't think the compiler is smart enough to put things into constant memory unless specifically requested. I think you should be able to just make the argument a host gtensor, and then that would end up being captured by copy into the lambda function. 

That's kinda what I remember, anyway -- it may be worth looking how the actual corresponding GENE kernels work. I kinda remember that even though the coefficients are in managed mem, they aren't actually accessed directly from the kernel but end up being passed by value. The difference being, though, that there it's using expressions not a lambda function, and so those are in fact scalars that can definitely be captured by value and end up being copied as they are passed as arguments to the actual kernel.

bd4: The messy bit here is that for the array syntax version, the coefficients need to be on host, since it constructs a gfunction with the coefficient and the gtensor_span. For the fused version, it's not possible (?) to use array syntax, so it's a launch kernel, and in that case the coefficients need to either be on device, or captured by value into the lambda. To make matters more interesting, the same routine is used for the host gtensor implementation, with a different wrapper.

For the purpose of benchmarking, I think it may be useful to have both fused and non-fused versions, to get a sense for how much difference it makes. I also want to keep the benchmark similar to what actually happens in GENE - this is why I do the awkward thing of using storage objects in the top level test routine and adapting them, rather than just using high level gtensor types everywhere. While capturing by value into the lambda is possibly more performant, the direct comparison with the hand written GPU kernel would be to use a device array for the coefficients.

I'm pretty sure GENE still uses the hand written CUDA/HIP kernel for this, not a GT port, even though we have experimented with performance of the GT port.

bd4:
I believe this is ready now - fixed the fused version, and changed to run both non-fused and fused. Code could be refactored to have less copy paste... but is good enough now I think.

germasch:
> I believe this is ready now - fixed the fused version, and changed to run both non-fused and fused. Code could be refactored to have less copy paste... but is good enough now I think.

Sounds fair enough ;)

PR #147, opened 2021-12-09
Pr/sycl gene compat

bd4:


include/gtensor/macros.h
```diff
@@ -100,6 +104,18 @@ inline void doHipCheck(hipError_t code, const char* file, int line)
...
+#ifndef NDEBUG
+#define gpuSyncIfEnabled()                                                     \
+  do {                                                                         \
+  } while (0)
+#else // NDEBUG defined
+#define gpuSyncIfEnabled()                                                     \
+  do {                                                                         \
+  } while (0)
+#endif // NDEBUG
+
```
germasch, line 118: I guess this could consolidated, assuming sync will always remain a no-op on sycl. I know too little about sycl at this point. I gather stuff will be submitted to a queue -- is that queue not executing stuff asynchronously? It seems there's likely still a need to sync somewhere, right?

germasch (approved):
Other than one question, LGTM.

include/gtensor/macros.h
bd4: I refactored the headers to this can be a real sync on SYCL

PR #148, opened 2021-12-21
sycl: use LP64, works with GPU backends

bd4:
Note that ILP64 is still required if using a CPU SYCL backend

### DIFF
--- a/.github/workflows/sycl.yml
+++ b/.github/workflows/sycl.yml
@@ -17,7 +17,8 @@ jobs:
       SYCL_DEVICE_FILTER: host
       DEBIAN_FRONTEND: noninteractive
       GTENSOR_TEST_EXCLUDE: test_fft test_reductions
-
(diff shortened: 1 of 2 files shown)


germasch (approved):
This looks fine with me, though I guess eventually there should probably be a way to support either lp64/ilp64 and correspondingly define an integer index type to go with that (which of course still wouldn't translate to Fortran directly).

bd4:
> This looks fine with me, though I guess eventually there should probably be a way to support either lp64/ilp64 and correspondingly define an integer index type to go with that (which of course still wouldn't translate to Fortran directly).

The issue is that the host (and probably opencl CPU) backends require ILP64. For the GPU backend, it doesn't matter which mkl lib is linked because the code is elsewhere, but the interface itself is currently 64bit index types. I updated it so it can be specified independently, via GTENSOR_DEVICE_SYCL_ILP64. This will give us maximum flexibility as the situation evolves. Note that the code is not yet check this value (because of the situation with GPU always using ILP64, this would be confusing - there is an interface value, and a linking / define MKL_ILP64 or not value).

PR #150, opened 2022-01-11
Pr/transform reduce

bd4:
Also adds `gt::sum_squares` helper that works on arbitrary expressions.

### DIFF
--- a/include/gtensor/reductions.h
+++ b/include/gtensor/reductions.h
@@ -8,6 +8,7 @@
 #endif
 
 #include <assert.h>
+#include <functional>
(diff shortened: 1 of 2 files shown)


bd4:
Working on host, cuda, and hip, but for some reason fails to compile with DPCPP and sycl backend. Not sure if this is a in Intel compiler or in the gtensor sycl implementation.

bd4:
sycl issue was stupid, missed a change to OutputType for init arg. Ready now I think.

germasch (approved):
LGTM.

Issue #153, opened 2022-01-28
improve kernel launch error checking


bd4:
Can add an error check in gtLaunchKernel macro, for HIP and CUDA backends. SYCL will raise an async exception on launch failure, which should already be handled by the global error handler.

bd4:
closed by #155

PR #155, opened 2022-01-29
Launch check

gmerlo:
This should address #153 printing out the blocks and threads if the launch fails.


germasch (approved):
Looks great ;)

bd4:
Thanks!

PR #159, opened 2022-02-17
Pr/flatten kernels

bd4:


germasch:
I think this 

> auto a = gt::flatten(b);

resulting in a copy is just regular C++ semantics. `auto&& a = gt::flatten(b);` is the way to go if one doesn't want that. 

Though there's more to think about as to whether `gt::flatten()` should or should not always return a view (in which case one gets reference semantics even with `auto`).

bd4:
I think it should always return a view - having it do a copy for 1d but not for 2d+ is confusing. Also reshape always returns a view, so doing an explicit flatten without the helper will also avoid the copy (I think... haven't tested that thoroughly).

bd4:
Slightly OT, but I think there is naming issue here - all the view methods in gview.h don't have anything indicating they are in a view family... there is gt::view, .view, then gt::reshape, gt::transpose, gt::swapaxes, all which have view/reference style semantics on assignment.

bd4:
According to my own documentation, gview actually has container semantics if it's wrapping a container. I think there was a good reason we decided on this... perhaps the `auto` vs `auto&&` subtlety in modern C++ is something we can't / shouldn't try to hide.

bd4:
I think the easy solution here is to let the 1d case use the general implementation, and only special case gscalar.

PR #160, opened 2022-02-23
First attempt to care for large launch dimensions

td-mpcdf:
As I need it quite urgent, I implemented a simple `get_launch_grid` function which in the 2d and 3d assignment case computes a grid and threadblock size which does not fail. 

For assignments of higher dimension the implementation does not change anything, nor does it change the kernel launches of arbitrary kernels.


bd4:
This has already been implemented and tested with GENE in PR #159

bd4:
#159 has been merged, and the cmake related changes are in #161

PR #162, opened 2022-02-24
sycl: fix parameter size limit in assign

bd4:
Fixes issue when RHS is a very complex expession and ends up being
bigger than 2k limit when passed as a SYCL kernel parameter.


germasch (approved):
LGTM

PR #166, opened 2022-03-04
small fixes related to include order

germasch:
I don't think I understand why these problems happened where gtensor is used in my other code (PSC), and not in any
of the tests, but well, they did appear there...

### DIFF
--- a/include/gtensor/gtensor.h
+++ b/include/gtensor/gtensor.h
@@ -13,6 +13,7 @@
 #include "gtensor_span.h"
 #include "gview.h"
 #include "operator.h"
+#include "space.h"
 
 namespace gt
 {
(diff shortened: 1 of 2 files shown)


include/gtensor/gtensor.h
```diff
@@ -13,6 +13,7 @@
 #include "gtensor_span.h"
 #include "gview.h"
 #include "operator.h"
+#include "space.h"
```
bd4, line 16: I don't think this is necessary for the fix, but doesn't hurt anything - listing all the dependent headers in gtensor.h, even if included earlier, has a documentation purpose I guess.

bd4 (approved):
I don't know how this has been working either - I even built GENE with it, but that was with SYCL backend. The correct way to have managed storage with caching is with managed_vector, just like for device vector. I will open a separate PR for that, and update the gt-blas SYCL backend to use it instead so it uses caching.

include/gtensor/gtensor.h
germasch: Yeah, thinking back I'm not sure -- it was complaining about the `storage_traits` being unknown, and I think this did make a difference when I put it in, but maybe not. In the end the main issue was that in this code, I need thrust as the backend, and so changing that to default off broke stuff.

PR #167, opened 2022-03-04
support thrust backend storage even if not used as default

germasch:
The change away from thrust as default backend actually caused me quite some breakage, and life gets easier if thrust backend storage is still supported even though when it's not used by default.

### DIFF
--- a/include/gtensor/space.h
+++ b/include/gtensor/space.h
@@ -11,8 +11,7 @@
 
 #include <vector>
 
-#ifdef GTENSOR_USE_THRUST
-#include <thrust/device_allocator.h>
+#ifdef GTENSOR_HAVE_THRUST
 #include <thrust/device_vector.h>
 #include <thrust/host_vector.h>
 #endif
@@ -95,7 +94,7 @@ struct storage_traits<gt::backend::device_storage<T, A>>
 
 #endif
 
-#ifdef GTENSOR_USE_THRUST
+#ifdef GTENSOR_HAVE_THRUST
 
 template <typename T, typename A>
 struct storage_traits<::thrust::host_vector<T, A>>
--- a/include/gtensor/allocator.h
+++ b/include/gtensor/allocator.h
@@ -6,7 +6,7 @@
 #include <memory>
 #include <utility>
 
-#ifdef GTENSOR_USE_THRUST
+#ifdef GTENSOR_HAVE_THRUST
 #include <thrust/device_ptr.h>
 #endif
 
@@ -26,7 +26,7 @@ bool is_valid(P p)
   return bool(p);
 }
 
-#ifdef GTENSOR_USE_THRUST
+#ifdef GTENSOR_HAVE_THRUST
 template <typename T>
 bool is_valid(::thrust::device_ptr<T> p)
 {
--- a/include/gtensor/defs.h
+++ b/include/gtensor/defs.h
@@ -4,6 +4,12 @@
 
 #include <cstddef>
 
+// This really should be defined by the build system, but it'll cause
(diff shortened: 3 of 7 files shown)


bd4:
So GTENSOR_USE_THRUST determines what gt::device_vector, gt::managed_vector, gt::device_storage, gt::managed_storage, and gt::gtensor_device use by default, but all the code around customizing thrust is still enabled? Do you use customized thrust directly in your code? The quick fix would be to enable GTENSOR_USE_THRUST in your build, then you get the old behavior back. I am fine having thrust extensions exposed even when it's not used as the storage backend, but needs to be tested thoroughly on both platforms, some of this stuff ends up being surprisingly fragile.

bd4:
Looks like just need to add -DGTENSOR_HAVE_THRUST to the GNU make build for examples. Then I think CI will pass.

include/gtensor/space.h
```diff
@@ -95,7 +94,11 @@ struct storage_traits<gt::backend::device_storage<T, A>>
 
 #endif
 
-#ifdef GTENSOR_USE_THRUST
+// FIXME, something like GTENSOR_HAVE_THRUST would be preferable
+#if defined(GTENSOR_DEVICE_CUDA) || defined(GTENSOR_DEVICE_HIP)
```
bd4, line 98: I think GTENSOR_HAVE_THRUST is equivalent to CUDA | HIP, based on what you've defined in the CMakeLists.txt and that we've decided to use it for reductions so it's required for HIP backend.

bd4 (approved):
Looks fine, once CI passes.

germasch:
> So GTENSOR_USE_THRUST determines what gt::device_vector, gt::managed_vector, gt::device_storage, gt::managed_storage, and gt::gtensor_device use by default, but all the code around customizing thrust is still enabled? Do you use customized thrust directly in your code? The quick fix would be to enable GTENSOR_USE_THRUST in your build, then you get the old behavior back. I am fine having thrust extensions exposed even when it's not used as the storage backend, but needs to be tested thoroughly on both platforms, some of this stuff ends up being surprisingly fragile.

Right. I do use thrust directly in my code, too, but primarily I used gtensor with a specified storage, that being `thrust::device_vector` with a custom allocator.

And yeah, i though `GTENSOR_USE_THRUST` would be the quick fix, and in fact it works. But my CI uses docker images, which use spack, which builds gtensor, and the gtensor package doesn't know about `GTENSOR_USE_THRUST`, so that would have been a huge pain to update the spack package, then the images, and what not, and I think it's generally worthwhile to support whatever is available, orthogonal to what is selected to be the default, which is how I'd like to think about `GTENSOR_USE_THRUST`.

germasch:
> Looks like just need to add -DGTENSOR_HAVE_THRUST to the GNU make build for examples. Then I think CI will pass.

Thanks. I guess that does point out how this will break GENE's POM build, so I changed things to `#define GTENSOR_HAVE_THRUST` in `defs.h` based on cuda/hip. That smells a little bit, but I think it'll avoid those other problems, and I'd rather not be the next person to break GENE.

include/gtensor/space.h
germasch: Yeah, as you can tell from this comment that I addressed elsewhere but forgot right there, I did at first go with the (cuda|hip) ifdef, which is perfectly fine, but I think it's better documented to have the thrust parts conditional on a macro that actually expresses whether thrust is available -- and that'd also allow to disable those parts if so desired at some point in the future.

[Related to this, we do depend the build on thrust in the hip case, but we don't in the cuda case, which isn't really right. But I guess since that works since thrust is always available with cuda, I'm not quite up to risking breakage by actually doing this "the right way".]

PR #168, opened 2022-03-11
sycl: use in-order queues, support tiles

bd4:


CMakeLists.txt
```diff
@@ -233,10 +235,19 @@ if ("sycl" IN_LIST GTENSOR_BUILD_DEVICES)
             "-device ${GTENSOR_DEVICE_SYCL_AOT_ARCHITECTURES}" >)
     endif()
 
+    if (GTENSOR_DEVICE_SYCL_PER_TILE)
+      target_compile_definitions(gtensor_sycl INTERFACE
+                                 GTENSOR_DEVICE_SYCL_PER_TILE=1)
+    else()
+      target_compile_definitions(gtensor_sycl INTERFACE
+                                 GTENSOR_DEVICE_SYCL_PER_TILE=0)
+    endif()
```
germasch, line 244: The usual pattern is to define the macro to 1 if enabled, but keep it entirely undefined if disabled.

include/gtensor/backend_sycl_device.h
```diff
@@ -20,6 +20,10 @@
 #include "CL/sycl/backend/opencl.hpp"
 #endif
 
+#ifndef GTENSOR_DEVICE_SYCL_PER_TILE
+#define GTENSOR_DEVICE_SYCL_PER_TILE 1
+#endif
+
```
germasch, line 26: Well, which won't work with this bit. But as of now I'm not sure what the goal is here.

include/gtensor/backend_sycl_device.h
```diff
@@ -145,6 +149,29 @@ inline uint32_t get_unique_device_id<cl::sycl::backend::ext_oneapi_level_zero>(
...
+inline std::vector<cl::sycl::device> get_devices_with_numa_sub(
+  const cl::sycl::platform& p)
+{
+  std::vector<cl::sycl::device> result;
+  for (auto root_dev : p.get_devices()) {
+    // Handle GPUs with multiple tiles, which can be partitioned based on numa
+    // domain
+    auto max_sub_devices =
+      root_dev.get_info<cl::sycl::info::device::partition_max_sub_devices>();
+    if (GTENSOR_DEVICE_SYCL_PER_TILE && max_sub_devices > 0) {
```
germasch, line 161: Well, I guess I understand now, but I don't like the direct use of a macro inside of the code like this. It seems having a static bool or something, and have it initialized to a particular default depending on the macro being set would be much cleaner.

germasch (changes requested):

include/gtensor/backend_sycl_device.h
bd4: The goal is to have the default be per tile devices, even if no macro is specified (e.g. with a GNU make build using gtensor as a subproject). I am not 100% a static bool will work, if this is included in multiple places, e.g. in cgtensor and in C++ cuda code. One alternative is to use ifdef, it's just messier to handle the if/else then.

bd4: I guess I could use `const bool device_per_tile`, and invert the define, so it's `GTENSOR_DEVICE_SYCL_NO_SUB_DEVICES` or something. cont shouldn't have any linkage issues. The inverted define is annoying, I'm not sure what to call it, but I guess it is more standard.

germasch: Maybe the first question is, is the single queue behavior even still needed? gtensor already has more than enough macro switches, which doesn't make it easier for a user to get the right combination and, e.g., caused the issues I had with my other code recently. And compile-time switches are (usually) worse in that one may not catch stuff unless built with the right option.

Generally speaking, a static bool will work, but it's not the ideal solution, either.

bd4: I want the switch in case there is a bug related to running on both tiles.

bd4: Went with the env var implementation, cleaner and more useful for debugging to change at runtime.

germasch (approved):

PR #173, opened 2022-04-08
Pr/bench 1d stencil

bd4:


germasch (approved):
Looks like you found a workaround for warming up ;)

PR #176, opened 2022-05-10
Pr/bandsolve invert

bd4:
Add routines for efficient batched linear solve using inverted matrices. Includes bandsolver routine for inverting a batch of LU factored matrices, and gpu gemm_batched wrappers around native CUDA/HIP/SYCL APIs. Note that the bandsolver routines will also work on non-banded routines - it will just behave like a normal dense solve, but not as efficient since it is serial per batch and rhs. The idea is to have alternatives to native getrs_batched routines, which have not been optimized yet on all GPU platforms for the sizes typically used by GENE.

### DIFF
--- a/include/gt-blas/bandsolver.h
+++ b/include/gt-blas/bandsolver.h
@@ -15,6 +15,15 @@ struct matrix_bandwidth
   int upper;
 };
 
+/**
+ * Calculate max bandwidth in a batch of square matrices.
+ *
+ * @param n size of each A_i
+ * @param d_Aarray Array of device pointers to input [A_i]
+ * @param lda leading distance of each A_i, >=n
+ * @param batchSize number of matrices [A_i] and [B_i] in batch
+ * @return matrix_bandwidth struct containing max upper and lower bandwidth
+ */
 template <typename T>
 inline matrix_bandwidth get_max_bandwidth(int n, T** d_Aarray, int lda,
                                           int batch_size)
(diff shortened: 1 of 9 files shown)


include/gt-blas/bandsolver.h
```diff
@@ -133,6 +160,126 @@ inline void getrs_banded_batched(int n, int nrhs, T** d_Aarray, int lda,
...
+
+      for (int i = 0; i < n; i++) {
+        B[i] = 0;
+      }
+      B[rhs] = T(1);
+
+      for (int i = 0; i < n; i++) {
+        tmp = B[i];
+        B[i] = B[piv[i] - 1];
+        B[piv[i] - 1] = tmp;
```
germasch, line 201: `std::swap`?

include/gt-blas/bandsolver.h
```diff
@@ -133,6 +160,126 @@ inline void getrs_banded_batched(int n, int nrhs, T** d_Aarray, int lda,
...
+ * @param d_Barray Array of device pointers to each output B_i
+ * @param ldb leading distance of each B_i, >=n
+ * @param batchSize number of matrices [A_i] and [B_i] in batch
+ * @param lbw max lower bandwidth of all [A_i]
+ * @param ubw max upper bandwidth of all [A_i]
+ */
+template <typename T>
+inline void invert_banded_batched(int n, T** d_Aarray, int lda,
+                                  index_t* d_PivotArray, T** d_Barray, int ldb,
+                                  int batchSize, int lbw, int ubw)
```
germasch, line 182: I'm not really a big fan of passing raw pointers without information about size and where the data lives. But I'm also not totally sure how this all fits together -- if this is done to match an existing library interface, then that's a good justification, but otherwise I'd rather have this a more C++/gtensor specific interface.

germasch (changes requested):
LGTM to me for the most part. Generally speaking, it wouldn't be a bad idea to use the description box in the PR for its intended purpose ;)

include/gt-blas/bandsolver.h
bd4: doesn't work on AMD gpu unfortunately

include/gt-blas/bandsolver.h
bd4: It is designed to be similar to getrf/getrs to simplify usage from Fortran. Adding high level C++ wrappers that take gtensor container objects would be a nice value add -- but those don't exist for getrf/rs yet either. The awkwardness here is the device array of device pointers to device memory thing - making them gtensor objects doesn't remove that, and arguably makes it a bit more opaque to set up. Maybe worth adding a nice high level interface at some point... but again the only use case I have now is calling from Fortran... so hard to justify now.

germasch: Ok, fair enough.

bd4: To expand a little - the typical flow is to set up input matrices A, space for pivot and result, then call `gt::blas::getrf_batched`. The resulting LU factored matrices in A and the pivot are then passed to `gt::blas::invert_banded_batched`, with output array B that is same size as A. Then `gt::blas::gemm_batched` can be used to batch solve one or more RHS vectors per batch, using the inverted matrices. Not obvious to me how to do in place invert. Agreed this is fragile - particularly knowing the required dims for B. The batched blas interfaces have been "fun" to work with, particularly oneMKL... and I am not solving that here. Just not motivated to solve it, since i have to call from Fortran anyway so it's a mess no matter what I do....

germasch (approved):

PR #177, opened 2022-05-13
Pr/test gene 6d

bd4:
Add tests for non-contiguous view assignment, and attempt to improve error handling on dimension mismatch. This is to aid in debugging the 0x2b errors on crusher.

### DIFF
--- a/include/gtensor/gfunction.h
+++ b/include/gtensor/gfunction.h
@@ -2,6 +2,8 @@
 #ifndef GTENSOR_GFUNCTION_H
 #define GTENSOR_GFUNCTION_H
 
+#include <cassert>
+
 #include "defs.h"
 #include "expression.h"
 #include "gscalar.h"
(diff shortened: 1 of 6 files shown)


include/gtensor/gfunction.h
```diff
@@ -2,6 +2,8 @@
 #ifndef GTENSOR_GFUNCTION_H
 #define GTENSOR_GFUNCTION_H
 
+#include "assert.h"
+
```
germasch, line 6: Is this the standard C header? If so, it should be included via `#include <assert.h>`, or really, via `#include <cassert>`. I think/hope it's not a header included with gtensor, since that'd obviously be confusing.

include/gtensor/gtensor_span.h
```diff
@@ -2,7 +2,7 @@
 #ifndef GTENSOR_GTENSOR_VIEW_H
 #define GTENSOR_GTENSOR_VIEW_H
 
-#include <cassert>
+#include <assert.h>
```
germasch, line 5: Well, I guess there must have been a reason to change from `cassert` (cuda device code?). Nevertheless, the one mentioned above should use angled brackets.

germasch (changes requested):
LGTM except one minor nitpick.

include/gtensor/gtensor_span.h
bd4: Device side assert was not working, even when NDEBUG is not set. I imagine cassert includes assert.h eventually, so I don't think it should make a difference... and appears not to have. I guess I will change it back. I still don't understand why device side assert is doing nothing.

bd4:
I think the shape check was disabled before we had the scalar specific assign specialization. I am going to test this branch with GENE, and make sure it works there; it does pass the unit tests.

bd4:
@germasch the assert issue is fixed now

germasch (approved):

PR #178, opened 2022-06-14
fix assign broadcast

bd4:
Assign check is too strict, can broadcast if RHS has 1 in any
dimension. See calc_index in strides.h.


germasch (approved):
This looks good to me. Thanks!

PR #179, opened 2022-06-20
sycl: add AOT options when linking CXX

bd4:
AOT (ahead of time compilation) is broken without this


germasch (approved):
LGTM, I guess...

PR #180, opened 2022-06-23
Pr/gemm tests

bd4:
- adds tests to help debug blas and solver issues for SYCL backend
- improve sycl gemm_batched backend, fixes ci running on host
- adds managed memory comparison to some benchmarks
- fix ROCm build issue with gcc toolchain
- fix is_device_address for cuda backend


bd4:
Managed memory comparison is not very nuanced - it just subs the device gtensor container for a managed mem backed one, and still does an explicit copy. It's a start though, and may show up obvious differences.

PR #181, opened 2022-07-06
Pr/getrs benchmark

uphoffc:
I was debugging getrs_banded_batched as I got wrong results in GENE. It turned out to be something completely different, but I want to share the improved benchmark code.
Improvements:
- Parameter or file input is selected at run-time instead of compile time. Use "./getrs file file_name.txt" for file input.
- Added residual checking routine to assess accuracy. See check() function.

### DIFF
--- a/benchmarks/getrs.cxx
+++ b/benchmarks/getrs.cxx
@@ -12,30 +12,163 @@
 
 #define NRUNS 10
 
-//#define READ_INPUT
+template <typename CT>
+struct test_problem
+{
+  gt::gtensor<CT, 3> h_Adata;
+  int lda;
+  gt::gtensor<CT, 3> h_Bdata;
+  int ldb;
+  gt::gtensor<gt::blas::index_t, 2> h_piv;
+};
 
 template <typename T>
-inline void read_carray(std::ifstream& f, int n, T* Adata)
-{
-  for (int i = 0; i < n; i++) {
-    // std::cout << i << " " << std::endl;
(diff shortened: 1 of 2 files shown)


bd4:
Thanks this looks great! I need to do some tests on JLSE to make sure it works the same for the Aurora test suite, then I think can be merged, been busy putting out other fires.

benchmarks/getrs.cxx
```diff
@@ -12,30 +12,163 @@
...
+        }
+      }
+      // Compute ||L tmp - rhs||_1 and ||rhs||_1
+      for (std::size_t i = 0; i < n; ++i) {
+        auto b_i = tmp(i);
+        for (std::size_t j = 0; j < i; ++j) {
+          b_i += tp.h_Adata(i, j, b) * tmp(j);
+        }
+        residual_L1 += std::abs(b_i - rhs(i));
+        b_L1 += std::abs(rhs(i));
```
bd4, line 98: These need to be changed to `gt::abs`, certain HIP/CUDA versions fail to build when using `std::abs` on `thrust::complex`

bd4 (changes requested):
This is very helpful, thank you! Just needs minor change std::abs -> gt::abs.

Note that the residual checks do pass for banded and invert solvers on CUDA / cuBLAS backend.

bd4:
Residual check also passes with SYCL backend on Gen9 with public oneapi release.

PR #187, opened 2022-07-20
storage: fix memory leak on move assignment

bd4:
For non thrust backend, anything that does a move assignment of an owning gtensor object will result in move assignment of the underlying gtensor_storage, which was not correctly deallocating the existing destination data and resulting in memory leaks.


gmerlo (approved):

germasch (commented):
Good catch! Maybe using `std::unique_ptr` for `data_` would be a way to fortify against this kind of bugs.

PR #188, opened 2022-07-21
Pr/blas getri

bd4:
Adds `gt::blas::getri_batched`, along with tests and benchmark, and refactors and improves stream handling. 

* refactor stream handling in gt-blas, using stream_view for set/get instead of gt::blas::stream_t
* bandsolver family of calls now take a blas handle, and respect the stream set on the handle
* reduction API now has an optional stream_view argument.
* C API still uses gt::blas::stream_t, to keep disruption to existing Fortran code to a minimum.

### DIFF
--- a/benchmarks/bench_getri.cxx
+++ b/benchmarks/bench_getri.cxx
@@ -119,7 +119,7 @@ static void BM_getri(benchmark::State& state)
   gt::copy(h_AAinvptr, d_AAinvptr);
   gt::synchronize();
 
-  gt::blas::handle_t* h = gt::blas::create();
+  gt::blas::handle_t h;
 
   gt::blas::getrf_batched(h, N, gt::raw_pointer_cast(d_Aptr.data()), N,
                           gt::raw_pointer_cast(d_piv.data()),
@@ -155,8 +155,6 @@ static void BM_getri(benchmark::State& state)
   for (auto _ : state) {
     fn();
   }
-
-  gt::blas::destroy(h);
 }
 
 // RealType, N, NBATC
--- a/benchmarks/bench_getri.cxx
+++ b/benchmarks/bench_getri.cxx
@@ -12,36 +12,7 @@
 
 using namespace gt::placeholders;
 
-//#define BENCH_GETRI_DEBUG
-
-template <typename Matrix>
-void debug_print_matrix(const char* name, Matrix&& M)
-{
-#ifdef BENCH_GETRI_DEBUG
-  int n = M.shape(0);
-  std::cerr << name << "\n";
-  for (int j = 0; j < n; j++) {
-    for (int i = 0; i < n; i++) {
-      std::cerr << M(i, j) << " ";
-    }
-    std::cerr << "\n";
-  }
-  std::cerr << std::endl;
-#endif
-}
-
-template <typename Vector>
-void debug_print_vector(const char* name, Vector&& V)
-{
-#ifdef BENCH_GETRI_DEBUG
-  int n = V.shape(0);
-  std::cerr << name << "\n";
-  for (int i = 0; i < n; i++) {
-    std::cerr << V(i) << " ";
-  }
-  std::cerr << std::endl;
-#endif
-}
+// #define BENCH_GETRI_DEBUG
 
 template <typename CT>
 auto make_test_matrix(int n, int bw, int batch_size, bool needs_pivot)
@@ -177,7 +148,7 @@ static void BM_getri(benchmark::State& state)
   check_close_identity_batch<R>(h_AAinv);
 
   for (int b = 0; b < NBATCH; b++) {
-    debug_print_matrix("A*Ainv", h_AAinv.view(_all, _all, b));
+    std::cout << b << " A*Ainv\n" << h_AAinv.view(_all, _all, b) << std::endl;
   }
 #endif
 
--- a/benchmarks/bench_getri.cxx
+++ b/benchmarks/bench_getri.cxx
@@ -1,0 +1,210 @@
+#include <iostream>
+#include <numeric>
+#include <string>
+
+#include <benchmark/benchmark.h>
+
+#include <gtensor/gtensor.h>
+
+#include <gt-blas/blas.h>
+
+#include "gt-bm.h"
+
+using namespace gt::placeholders;
+
+//#define BENCH_GETRI_DEBUG
+
+template <typename Matrix>
+void debug_print_matrix(const char* name, Matrix&& M)
+{
+#ifdef BENCH_GETRI_DEBUG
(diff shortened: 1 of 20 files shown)


benchmarks/bench_getri.cxx
```diff
@@ -0,0 +1,204 @@
...
+#ifdef BENCH_GETRI_DEBUG
+  int n = V.shape(0);
+  std::cerr << name << "\n";
+  for (int i = 0; i < n; i++) {
+    std::cerr << V(i) << " ";
+  }
+  std::cerr << std::endl;
+#endif
+}
+
```
germasch, line 45: FWIW, I think we do have code that allows `std::cout << M` etc. If that doesn't work / isn't helpful, it may be worth looking at fixing it. (Also, it might be useful to have something that dumps an xtensor into, e.g., a numpy-format file)

include/gt-blas/bandsolver.h
```diff
@@ -255,13 +274,15 @@ inline void invert_banded_batched(int n, T** d_Aarray, int lda,
  * @param batchSize number of matrices [A^-1_i], [B_i], [C_i] in batch
  */
 template <typename T>
-inline void solve_inverted_batched(int n, int nrhs, T** d_Aarray, int lda,
-                                   T** d_Barray, int ldb, T** d_Carray, int ldc,
-                                   int batchSize)
+inline void solve_inverted_batched(handle_t* h, int n, int nrhs, T** d_Aarray,
+                                   int lda, T** d_Barray, int ldb, T** d_Carray,
+                                   int ldc, int batchSize)
```
germasch, line 279: Why is `h` passed as a pointer? I earlier got the impression that handles themselves are easily copied (preserving reference semantics, since usually the internal of a handle is just a pointer, anyway).

Even if there is a reason to avoid a copy, I'd think passing the handle by reference would be preferred over the explicit pointer. That is, unless `h` could be `nullptr`, but I don't think that's the case here.

src/cgtblas.cxx
```diff
@@ -91,12 +91,13 @@ void gtblas_destroy()
...
-  gt::blas::set_stream(g_handle, stream_id);
+  gt::blas::set_stream(g_handle, gt::stream_view{stream_id});
 }
 
 void gtblas_get_stream(gt::blas::stream_t* stream_id)
 {
-  gt::blas::get_stream(g_handle, stream_id);
+  auto sview = gt::blas::get_stream(g_handle);
+  *stream_id = sview.get_backend_stream();
 }
```
germasch, line 101: Is there a reason to not have this function return its result instead? (This is a question about pre-existing behavior)

tests/test_lapack.cxx
```diff
@@ -593,3 +593,103 @@ TEST(lapack, zgetrs_batch_managed)
...
+  expect_complex_near(h_C(1, 0, 1), T(0.04, 0.28));
+  expect_complex_near(h_C(2, 0, 1), T(0.52, -0.36));
+  // second batch, second solution col
+  expect_complex_near(h_C(0, 1, 1), T(-0.04, -0.28));
+  expect_complex_near(h_C(1, 1, 1), T(0.016, -0.088));
+  expect_complex_near(h_C(2, 1, 1), T(-0.092, 0.256));
+  // second batch, third solution col
+  expect_complex_near(h_C(0, 2, 1), T(0.07, -0.01));
+  expect_complex_near(h_C(1, 2, 1), T(-0.028, -0.096));
+  expect_complex_near(h_C(2, 2, 1), T(0.036, 0.052));
```
germasch, line 674: Not really important, but in cases like this, wouldn't it be possible to define a reference solution and than do something like `expect_near(h_C, ref_C);`?

germasch (changes requested):
Looks good to me with some minor comments.

It does generally somewhat bother me, though, that there is this rather huge amount of code for something that doesn't seem like it should be that complex...

include/gt-blas/bandsolver.h
bd4: The only advantage I see of a pointer is that it allows more flexibility with Fortran interop and flexible scope. We are not using that now though - the C lib uses a global handle and the handle never makes it into Fortran land. I'm not opposed to changing it to use reference.

src/cgtblas.cxx
bd4: Only reason is less work in changing Fortran interface and call sites. Re original design, it was modeled after cublas, which also involved less changes to the existing Fortran interface call cublas_set_stream etc.

tests/test_lapack.cxx
bd4: array near already exists actually - it usually comes down to whether I copy/pasted one of the blas tests from an older test before I wrote that or a newer one. If I do heavy modifications to a test, I usually rewrite it, but other times feels not worth the effort.

include/gt-blas/bandsolver.h
germasch: This is a templated function, ie., not one which could possibly be called directly from Fortran. I think passing as `handle_t&` would be just fine, if we want to avoid any potential complications about whether a copy of `handle_t` has reference semantics or not. I think the modern C++ is basically to avoid raw pointers wherever possible, and if one has a choice of using reference vs pointer, a reference should be used when it's clear that there's always an object present, as opposed to a pointer, which would be used if `nullptr` is an expected (and handled) choice.

src/cgtblas.cxx
germasch: It's not something that needs to be changed (as I said, it's not something which was changed in this PR). But generally speaking, I think in C++ in particular it is preferable to return things that should be returned (in particular given return value optimization etc it's usually not more expensive, and there's pair/tuple etc for multiple values). And I would agree that having two different function signatures for C++ vs C wrappers isn't necessarily making life easier.

include/gt-blas/bandsolver.h
bd4: I think handle_t should also be a class that manages the underlying e.g. cublasHandle_t (calling create and destroy as needed). Then the instances can still be passed by reference, and if non lexical scope is really needed, caller can use unique_ptr. Fortran interface can still be whatever.

bd4:
> Looks good to me with some minor comments.
> 
> It does generally somewhat bother me, though, that there is this rather huge amount of code for something that doesn't seem like it should be that complex...

Wrapping APIs with slight differences (or in the case of oneMKL, significant differences) is a mess. I tried to use templates and in some case C preprocessor to make it less bad, but only so much one can do without going down the path of code generation, and I would argue that adds new problems and isn't worth it for this. There are efforts in C++ standardization for linear algebra, hopefully that eventually leads to something that makes the situation better.

germasch (approved):
LGTM!

PR #189, opened 2022-07-26
stream: refactor stream view and owning classes

bd4:
Use common stream classes. The existing code relies on each backend defining them correctly, with no help from the compiler.

Note that the hip and cuda backends still use a subclass that adds get_execution_policy - this is used by thrust reductions. Perhaps there is a more elegant way of handling this. It could be part of the interface, and sycl has a dummy implementation. Or sycl could have an implementation using oneDPL, but that makes us even more dependent on oneAPI vs SYCL standard.

### DIFF
--- a/include/gtensor/backend_cuda.h
+++ b/include/gtensor/backend_cuda.h
@@ -204,50 +204,59 @@ inline bool is_device_address(const Ptr p)
           attr.type == cudaMemoryTypeManaged);
 }
 
+namespace stream_interface
+{
+
+template <>
+inline cudaStream_t create<cudaStream_t>()
+{
+  cudaStream_t s;
+  gtGpuCheck(cudaStreamCreate(&s));
+  return s;
+}
+
+template <>
+inline cudaStream_t get_default<cudaStream_t>()
+{
+  return nullptr;
+}
+
+template <>
+inline void destroy<cudaStream_t>(cudaStream_t s)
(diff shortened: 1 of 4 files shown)


germasch (approved):
I'm kinda not involved enough with this to have strong opinions. I'm not sure that CRTP is worth it here -- it does help make sure one gets a consistent interface, but there isn't really much code that's being shared. I think what we really want is something concept-like, which of course isn't in C++14, but one could instead use a set of tests to make sure that certain semantics are satisfied by the various backend stream and stream_view classes.

One thing that irks me somewhat is that the owning stream class internally stores the thing as a view, which is decidedly not the semantics the class itself provides, but I suppose that this did make life easier.

bd4:
> I'm not sure that CRTP is worth it here
I tried that initially, and ended up with something I thought was pretty ugly, and landed on this compromise. Concepts would definitely make this much cleaner.

Re the view, I could have it hold the stream_ and just call get_view().X each time - this way I am creating less view object, most likely a pointless optimization. The main thing is to avoid re-implementing all the getters that are implemented in the view.

I'm not totally happy with where it is now, but it's marginally better than what was there before, where the classes where the same by inspection only with no compiler checking. Could be convinced to iterate more, this isn't very high priority, just a general clean up in line with what I was doing for gt-blas.

Issue #191, opened 2022-08-05
consistent size and index types


bd4:
Currently gtensor uses `int` based `shape_type`, while low level indexing like `gtensor_storage` index operator are `gt::size_type`, which is an alias for `std::size_t` (typically unsigned long). Furthermore, the `calc_size(shape)` helper returns `size_type`, but there are places in the code which index over a total size with int or cast it to int.

To make matters more interesting, the `dim3` type in CUDA and HIP is `uint32_t` (actually unsigned in for CUDA but that is generally uint32_t). It seems like we should at least be consistent with dim3 types?

bd4:
FWIW, the SYCL spec landed on using `size_t` for all index types, and `int` for dimensions: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:item.class

The fact that the gpu blas APIs typically use int indexes makes things fun here. Bottom line, we don't want silent failures if overflow happens, even in release builds.

germasch:
I guess I'll agree that while there was some intention behind the different size for multi-d index vs 1-d index, it's really not possible to not do this in a way that doesn't cause potential subtle overflow bugs.

So probably what we should do is to use `<class>::size_type` consistently everywhere, get rid of `int`, with the former being defaulted to `gt::size_type` which in itself should probably default to `std::size_t` (but one could argue about `uint`).

Maybe actually `ssize_t` would be better, since there are times where negative indices may make some sense. I think this need some more thinking/investigation...

bd4:
yes this is subtle. As a start, I am using size_t for index calculations inside kernels and for SYCL and whenever doing interations over container.size() or calc_size(shape). This involves no change to the external interface.

bd4:
There are potentially 3 types at play here:

1. size_type - for addressing in to memory
2. index_type, backend - what the underlying GPU lib uses, e.g. for dim3 in CUDA (uint32_t)
3. index_type, gt high level - currently int, as part of shape_type, not convinced this should be different from (2)

I'm inclined to make gt::index_type a standard that should be used e.g. for `GT_LAMBDA (gt::index_type i)`, and then we have knobs to change this and aren't stuck with a bunch of code with hard coded ints. I am also inclined to change the default to uint32_t on HIP/CUDA and to size_t for SYCL, to match the backend.

bd4:
[...]

PR #192, opened 2022-08-05
indexing: use size_type for flat index, nd kernel

bd4:
Does not change external interface (shape_type is still int sarray).
This should make sure there is no int overflow in internal index
calculations. Note that reshape could still overflow, but the
assert that the sizes stayed the same in the implementation should
at least catch it in debug builds.

### DIFF
--- a/include/gtensor/assign.h
+++ b/include/gtensor/assign.h
@@ -344,10 +344,10 @@ struct assigner<6, space::device>
 #else // not defined GTENSOR_PER_DIM_KERNELS
 
 template <typename Elhs, typename Erhs, size_type N>
-__global__ void kernel_assign_N(Elhs lhs, Erhs rhs, int size,
+__global__ void kernel_assign_N(Elhs lhs, Erhs rhs, size_type size,
                                 gt::shape_type<N> strides)
 {
-  int i = threadIdx.x + blockIdx.x * blockDim.x;
+  size_type i = threadIdx.x + blockIdx.x * blockDim.x;
 
   if (i < size) {
     auto idx = unravel(i, strides);
@@ -361,9 +361,12 @@ struct assigner<N, space::device>
   template <typename E1, typename E2>
   static void run(E1& lhs, const E2& rhs, stream_view stream)
   {
-    int size = int(calc_size(lhs.shape()));
+    auto size = calc_size(lhs.shape());
     auto strides = calc_strides(lhs.shape());
-    auto block_size = std::min(size, BS_LINEAR);
+    unsigned int block_size = BS_LINEAR;
+    if (block_size > size) {
+      block_size = static_cast<unsigned int>(size);
+    }
 
     dim3 numThreads(block_size);
     dim3 numBlocks(gt::div_ceil(size, block_size));
@@ -399,7 +402,7 @@ struct assigner<1, space::device>
       using rtype = decltype(k_rhs);
       using kname = gt::backend::sycl::Assign1<E1, E2, ltype, rtype>;
       cgh.parallel_for<kname>(range, [=](sycl::item<1> item) {
-        int i = item.get_id();
+        auto i = item.get_id();
         k_lhs(i) = k_rhs(i);
       });
     });
@@ -425,8 +428,8 @@ struct assigner<2, space::device>
       using rtype = decltype(k_rhs);
       using kname = gt::backend::sycl::Assign2<E1, E2, ltype, rtype>;
       cgh.parallel_for<kname>(range, [=](sycl::item<2> item) {
-        int i = item.get_id(1);
-        int j = item.get_id(0);
+        auto i = item.get_id(1);
+        auto j = item.get_id(0);
         k_lhs(i, j) = k_rhs(i, j);
       });
     });
@@ -452,9 +455,9 @@ struct assigner<3, space::device>
       using rtype = decltype(k_rhs);
       using kname = gt::backend::sycl::Assign3<E1, E2, ltype, rtype>;
       cgh.parallel_for<kname>(range, [=](sycl::item<3> item) {
-        int i = item.get_id(2);
-        int j = item.get_id(1);
-        int k = item.get_id(0);
+        auto i = item.get_id(2);
+        auto j = item.get_id(1);
+        auto k = item.get_id(0);
         k_lhs(i, j, k) = k_rhs(i, j, k);
       });
     });
@@ -515,7 +518,7 @@ template <typename S>
 inline void valid_assign_broadcast_or_throw(const S& lhs_shape,
                                             const S& rhs_shape)
 {
-  for (int i = 0; i < lhs_shape.size(); i++) {
+  for (size_type i = 0; i < lhs_shape.size(); i++) {
     if (lhs_shape[i] != rhs_shape[i] && rhs_shape[i] != 1) {
       throw std::runtime_error("cannot assign lhs = " + to_string(lhs_shape) +
                                " rhs = " + to_string(rhs_shape) + "\n");
--- a/include/gtensor/assign.h
+++ b/include/gtensor/assign.h
@@ -347,7 +347,8 @@ template <typename Elhs, typename Erhs, size_type N>
 __global__ void kernel_assign_N(Elhs lhs, Erhs rhs, size_type size,
                                 gt::shape_type<N> strides)
 {
-  size_type i = threadIdx.x + blockIdx.x * blockDim.x;
+  // workaround ROCm 5.2.0 compiler bug
+  size_type i = threadIdx.x + static_cast<size_type>(blockIdx.x) * blockDim.x;
 
   if (i < size) {
     auto idx = unravel(i, strides);
--- a/include/gtensor/gtensor.h
+++ b/include/gtensor/gtensor.h
@@ -312,7 +312,8 @@ __global__ void kernel_launch(gt::shape_type<6> shape, F f)
 template <typename F, size_type N>
 __global__ void kernel_launch_N(F f, size_type size, gt::shape_type<N> strides)
 {
-  size_type i = threadIdx.x + blockIdx.x * blockDim.x;
+  // workaround ROCm 5.2.0 compiler bug
+  size_type i = threadIdx.x + static_cast<size_type>(blockIdx.x) * blockDim.x;
 
   if (i < size) {
     auto idx = unravel(i, strides);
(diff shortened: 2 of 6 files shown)


bd4:
Note that using size_type for index calculations could have performance impact.

include/gtensor/assign.h
```diff
@@ -138,7 +138,7 @@ struct assigner<6, space::host>
 template <typename Elhs, typename Erhs>
 __global__ void kernel_assign_1(Elhs lhs, Erhs rhs)
 {
-  int i = threadIdx.x + blockIdx.x * blockDim.x;
+  size_type i = threadIdx.x + blockIdx.x * blockDim.x;
```
germasch, line 141: I guess I'm not totally convinced of this, since likely current hardware doesn't support a configuration with that many threads, ie., it cannot possibly happen. Also, in particular in this case, since that index is then passed to `operator()`, it'd be truncated at that point, but that then can also not happen.

In the end, I don't particularly mind this, but it's something to be revisited. In particular, it'd probably be smart to not launch 100 million threads for an array of size 100 million anyway, but rather launch something more reasonable and then have a loop that catches all indices, I think that generally is also better for performance.

include/gtensor/assign.h
```diff
@@ -148,8 +148,8 @@ __global__ void kernel_assign_1(Elhs lhs, Erhs rhs)
 template <typename Elhs, typename Erhs>
 __global__ void kernel_assign_2(Elhs lhs, Erhs rhs)
 {
-  int i = threadIdx.x + blockIdx.x * BS_X;
-  int j = threadIdx.y + blockIdx.y * BS_Y;
+  size_type i = threadIdx.x + blockIdx.x * BS_X;
+  size_type j = threadIdx.y + blockIdx.y * BS_Y;
```
germasch, line 152: I'm at this point not sure whether you're changing `operator(int{, int, ...)`, but I don't think we should. So since i and j are used for that, we shouldn't pretend they could be that large (and at some point one really might take a performance hit from doing more 64 bit calcs/registers than necessary).

include/gtensor/assign.h
```diff
@@ -344,7 +344,7 @@ template <typename Elhs, typename Erhs, size_type N>
 __global__ void kernel_assign_N(Elhs lhs, Erhs rhs, int size,
                                 gt::shape_type<N> strides)
 {
-  int i = threadIdx.x + blockIdx.x * blockDim.x;
+  size_type i = threadIdx.x + blockIdx.x * blockDim.x;
```
germasch, line 347: This one I have less concerns about, since the result gets passed to unravel, which takes a size type.

I guess the other comment on all of these is that they look like they may support a > 32 bit space, but they're not actually correct for that, since the r.h.s. will still be calculated in uint32_t (and possibly overflow), only the result is then extended to 64 bit.

include/gtensor/assign.h
```diff
@@ -422,8 +422,8 @@ struct assigner<2, space::device>
       using rtype = decltype(k_rhs);
       using kname = gt::backend::sycl::Assign2<E1, E2, ltype, rtype>;
       cgh.parallel_for<kname>(range, [=](sycl::item<2> item) {
-        int i = item.get_id(1);
-        int j = item.get_id(0);
+        size_type i = item.get_id(1);
+        size_type j = item.get_id(0);
```
germasch, line 426: Again, if operator() still takes int and not size_type, this makes it look like we support something we don't, and possibly makes it more expensive. I suppose `auto` would be just fine here.

include/gtensor/assign.h
```diff
@@ -512,7 +512,7 @@ template <typename S>
 inline void valid_assign_broadcast_or_throw(const S& lhs_shape,
                                             const S& rhs_shape)
 {
-  for (int i = 0; i < lhs_shape.size(); i++) {
+  for (size_type i = 0; i < lhs_shape.size(); i++) {
```
germasch, line 515: 👍

include/gtensor/gtensor.h
```diff
@@ -252,12 +252,12 @@ __global__ void kernel_launch(gt::shape_type<3> shape, F f)
...
-  int i = threadIdx.x + blockIdx.x * BS_X;
-  int j = threadIdx.y + blockIdx.y * BS_Y;
+  size_type i = threadIdx.x + blockIdx.x * BS_X;
+  size_type j = threadIdx.y + blockIdx.y * BS_Y;
   int b = blockIdx.z;
-  int l = b / shape[2];
+  size_type l = b / shape[2];
   b -= l * shape[2];
-  int k = b;
+  size_type k = b;
```
germasch, line 260: These calculations are likely expensive as-is already, and extending them to 64 bit for no benefit doesn't seem worth it -- these are basically hardware-specific sections of the code which I think are ok to be tuned to the particular hardware capabilities (but comments about it wouldn't hurt).

include/gtensor/gtensor_storage.h
```diff
@@ -216,7 +216,7 @@ bool operator==(const gtensor_storage<T, A, O>& v1,
   auto&& h2 = host_mirror(v2);
   copy(v1, h1);
   copy(v2, h2);
-  for (int i = 0; i < h1.size(); i++) {
+  for (size_type i = 0; i < h1.size(); i++) {
```
germasch, line 219: 👍

germasch (changes requested):

include/gtensor/assign.h
bd4: `operator()` is actually templated, so will use whatever type it's passed. I think that this approach is too heavy handed in any case - my thinking now is that we should have size_type and / or index_t which are used consistently and can be configured with defines - that seems like the header-only library way of doing things.

include/gtensor/assign.h
bd4: operator takes whatever you give it, if I am reading the code correctly. So I think use size_type for sycl makes sense. And actually yes, `auto` would be better, we just inherit what sycl uses then.

include/gtensor/assign.h
bd4: I am inclined to use size_type at least for the nd implementations, because it's already using unravel. but it could still have performance impact...

bd4:
Related - I think using size_type for the dimension is overkill, but I don't know that it makes any difference, e.g. with compile time. SYCL uses int for dimensions, which surprises me a little, why not unsigned int?

bd4:
I think the first commit should be uncontroversial now. The other two are a step towards making it so we have more control over the index_type, but is maybe still not fully baked...

bd4:
Related to #191

include/gtensor/assign.h
germasch: I think it's probably too difficult to try to have distinct index_t, size_t and still get something consistent. E.g., if one flattens an array, one would expect `x_flat(i)` to be the same as `x.data()[i]`. So I think it'd be easier to say "we're using all 32-bit" vs "all 64-bit" and not try to be smart, as I was kinda trying to be. In addition, I still think in particular gpus may benefit from index computations never involving 64 bit.
The other part to this is what to do about signedness, though I guess one way is to just keep `operator()` templated and essentially leave it to the user. I can't quite wrap around my head of where signed values actually make sense, definitely not for regular accesses. xtensor has `.periodic(i, ...)`, where negative indices are meaningful, but I don't think we have that yet at all. For slices, negative indices are meaningful as well, but that should have its own distinct type. I guess part of this question is, what are we (ab-)using `shape_type` for?

bd4:
Force pushed just the commit changing internal use of size_type when dealing with .size and unravel, which I think is uncontroversial. The size_type v index_type thing is hopefully unnecessary, and needs more discussion

include/gtensor/assign.h
bd4: Removed from this pr pending further discusion

include/gtensor/assign.h
bd4: We use `shape_type` for strides. I can't think of any place it's used currently where it needs to take a negative value. I am leaning towards uint32_t as default size_type used by shape_type, maybe size_t for SYCL, with ability to change it with cmake options. Can deal with the reverse indexing e.g. in bandsolver as a special case. But needs more discussion, will be a separate PR when we decide.

include/gtensor/assign.h
```diff
@@ -341,10 +341,10 @@ struct assigner<6, space::device>
 #else // not defined GTENSOR_PER_DIM_KERNELS
 
 template <typename Elhs, typename Erhs, size_type N>
-__global__ void kernel_assign_N(Elhs lhs, Erhs rhs, int size,
+__global__ void kernel_assign_N(Elhs lhs, Erhs rhs, size_type size,
                                 gt::shape_type<N> strides)
 {
-  int i = threadIdx.x + blockIdx.x * blockDim.x;
+  size_type i = threadIdx.x + blockIdx.x * blockDim.x;
```
germasch, line 347: I still don't like this, since the rhs is `uint32_t` anyway. I think it'd be somewhat consistent to make the rhs `uint32_t` as well, or `auto` would work here as well.

include/gtensor/gtensor.h
```diff
@@ -303,9 +303,9 @@ __global__ void kernel_launch(gt::shape_type<6> shape, F f)
 #else // not GTENSOR_PER_DIM_KERNELS
 
 template <typename F, size_type N>
-__global__ void kernel_launch_N(F f, int size, gt::shape_type<N> strides)
+__global__ void kernel_launch_N(F f, size_type size, gt::shape_type<N> strides)
 {
-  int i = threadIdx.x + blockIdx.x * blockDim.x;
+  size_type i = threadIdx.x + blockIdx.x * blockDim.x;
```
germasch, line 308: same as above

germasch (changes requested):
Looks good with one issue remaining (repeated in two places).

include/gtensor/assign.h
germasch: I think you may have missed undoing this change to `size_type`, but did push commits from your other PR by mistake?

bd4: I left the size_type change in on purpose for the nd kernels, where it's immediately passed to unravel as size_t anyway. I just tested and this alone does not workaround the HIP compiler bug - casting blockIdx.x to size_type does. I think it is worth doing until we do a deeper refactor.

germasch (approved):
LGTM. I guess I'm not holding my breath for an immediate fix from AMD, so workaround it is.

PR #193, opened 2022-08-06
fix host reductions for HIP and CUDA

bd4:
Stream execution policy is not designed for use with host_vector


germasch (approved):
LGTM, but I guess the CI says some tests need to be updated for this.

bd4:
We could have the CI run the host side of CUDA tests after building, but would require some ifdef hackery to make that work, or some scripting around gtest-filter args.

PR #194, opened 2022-08-06
fix synchronization of gtensor ctor with thrust hip backend

bd4:


include/gtensor/gtensor.h
```diff
@@ -93,7 +93,14 @@ class gtensor_container : public gcontainer<gtensor_container<EC, N>>
...
   : base_type(shape, calc_strides(shape)), storage_(calc_size(shape))
-{}
+{
+#if defined(GTENSOR_USE_THRUST) && defined(GTENSOR_DEVICE_HIP)
+  // NOTE: rocThrust backend appears to not synchronize uninitialized zero fill
+  // by default, which can lead to suprising behavior where data is zeroed after
+  // constructor is run.
+  gt::synchronize();
+#endif
+}
```
germasch, line 103: 👍

include/gtensor/gtensor.h
```diff
@@ -102,6 +109,8 @@ inline gtensor_container<T, N>::gtensor_container(const shape_type& shape,
   : base_type(shape, calc_strides(shape)), storage_(calc_size(shape))
 {
   this->fill(fill_value);
+  // constructors should always be syncrhonous
+  gt::synchronize();
```
germasch, line 113: I don't really see a good reason why? Everything else can (potentially) be asynchronous, so what's special about constructors? I think the only exception is the default ctor, which doesn't actually touch anything, just allocates, so it's unexpected that one would have to wait for anything async, as is (was) the case with rocthrust.

Put differently, I think it's perfectly fine for `zeros_like` to launch an async kernel and one should expect that. Just `empty_like` really shouldn't, and if it does anyway, it should be hidden by the sync that you added.

germasch (changes requested):

include/gtensor/gtensor.h
bd4: I was thinking the ctor version would be sync and zeros_like / full would be async. But they call the ctor, so that doens't work, and perhaps is a more confusing place to draw the line than uninit is sync (because all it does is alloc) and everything else is async. I changed it.

germasch (approved):

PR #195, opened 2022-08-10
Bugfix/stream capi sycl

uphoffc:
In the SYCL variant of gtblas_get_stream the call to get_backend_stream() returns a sycl::queue&. But reference types are a C++ feature and not interoperable with C / Fortran. Here, I change stream_t to sycl::queue* for easy interopability. In order to distinguish between the cuda/hip and SYCL code-paths the detail::(get|set)_stream methods are introduced.

### DIFF
--- a/include/gt-blas/backend/sycl.h
+++ b/include/gt-blas/backend/sycl.h
@@ -13,7 +13,7 @@ namespace blas
 // ======================================================================
 // types aliases
 
-using stream_t = cl::sycl::queue;
+using stream_t = cl::sycl::queue*;
 using index_t = std::int64_t;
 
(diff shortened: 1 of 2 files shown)


bd4 (approved):
I think this is confusing but that is mostly because of existing design choices and it also fixes a major bug with gt-blas and the SYCL backend, so I think we should merge it and re-think how GENE and other apps are handling streams as a future step.

bd4:
Tested working as fix for segfault with GENE sycl backend.

germasch (approved):
LGTM, though I think we should eventually aim to have a more unified abstraction between the different backends.

PR #196, opened 2022-08-10
cmake: make SYCL backend builds  work with .cu

bd4:
Support for codes porting cuda code that use the .cu extension.
This works for CUDA backend and HIP backend, but requires this
extra option to work with SYCL backend and Intel dpcpp


germasch (approved):
Looks fine to me. I guess overall the best thing would be to try to avoid `.cu` files in the first place in the context of gtensor, since we're trying to abstract from whatever backend is in use.

PR #198, opened 2022-08-14
config: GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT opt

bd4:
Set to 'coarse' (default), 'device', or 'fine'. Note that this is
a change to default behavior - previous default was 'fine'.

### DIFF
--- a/include/gtensor/device_backend.h
+++ b/include/gtensor/device_backend.h
@@ -27,6 +27,50 @@ namespace backend
 {
 
 // ======================================================================
+// library wide configuration
+
+#ifndef GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT
+#define GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT "managed"
+#endif
+
+namespace config
+{
+
+class GtensorConfig
+{
+public:
+  GtensorConfig() : managed_memory_type_(GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT) {}
+
+  void set_managed_memory_type(const char* mtype)
+  {
+    managed_memory_type_ = mtype;
+  }
+
+  auto get_managed_memory_type() { return managed_memory_type_; }
+
+private:
+  const char* managed_memory_type_;
+};
+
+inline GtensorConfig& get_config_instance()
+{
+  static GtensorConfig config;
+  return config;
+}
+
+} // namespace config
+
+inline void set_managed_memory_type(const char* mtype)
+{
+  config::get_config_instance().set_managed_memory_type(mtype);
+}
+
+inline auto get_managed_memory_type()
+{
+  return config::get_config_instance().get_managed_memory_type();
+}
+
+// ======================================================================
 // stream interface
 
 namespace stream_interface
--- a/include/gtensor/device_backend.h
+++ b/include/gtensor/device_backend.h
@@ -30,7 +30,7 @@ namespace backend
 // library wide configuration
 
 #ifndef GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT
-#define GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT managed_memory_type::managed
+#define GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT managed
 #endif
 
 #ifdef GTENSOR_DEVICE_HIP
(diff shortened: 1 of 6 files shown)


germasch (changes requested):
I think macros are useful to be able to set a default without having to change add code into, e.g., GENE. But I think a nicer approach would be some kind of `gt::setManagedMemoryType()` or something that could be called programmatically. That might be useful to actually test/benchmark the various types without having to reconfigure/rebuild the whole thing, but most of all it means that all paths would always be compiled, which I think is one of the main drawbacks of ifdefs.

Again, the default could be chosen by a macro (and in that case, I think one can make a single macro work that takes values like `fine`, `coarse`, etc) Obviously one would have to think about what to do for non-hip backends (probably, ignore, except `device`, which makes sense on cuda, too).

bd4:
Re a library global setter, this would need to use the local static var hack to keep global state in header only library. This is what the SYCL backend uses to keep track of the "global" queue objects.

device/managed both make sense for cuda and sycl, at least for machines which have transparent host access to device mem (at the very least this is true for intel integrated gpu, e.g. laptop, and summit for cuda).

GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT = (managed, device, managed_course, managed_fine). On AMD, managed is an alias for managed_coarse if ROCm 5+, otherwise alias for managed_fine. Does that sound reasonable? I didn't realize one could do string compare in macros? Hopefully can drop support for ROCm < 5 in a few months, just need to make sure CI machines use it

bd4:
Interestingly, crusher is the only machine so far where transparent access to device memory truly works. Assigning it on Gen9 Intel integrated gpu, and on summit, results in segfault.

include/gtensor/device_backend.h
```diff
@@ -26,6 +26,60 @@ namespace gt
 namespace backend
 {
 
+// ======================================================================
+// library wide configuration
+
+#ifndef GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT
+#define GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT ManagedMemoryType::Managed
```
germasch, line 33: I think I would skip the namespace/enum in this define, it makes it awkward to use: `-DGTENSOR_MANAGED_MEMORY_TYPE_DEFAULT=Managed` is nicer than `GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT=ManagedMemoryType::Managed`, in particular in the case where someone might not be using cmake to take care of it.

include/gtensor/device_backend.h
```diff
@@ -26,6 +26,60 @@ namespace gt
...
+#endif
+
+#ifdef GTENSOR_DEVICE_HIP
+#if HIP_VERSION_MAJOR >= 5
+enum class ManagedMemoryType { Managed, Device, ManagedFine, ManagedCoarse };
```
germasch, line 38: Sorry to be nitpicky, but generally gtensor doesn't use camel case, but lower case with underscores, so I think we should stick with that.

include/gtensor/device_backend.h
```diff
@@ -26,6 +26,60 @@ namespace gt
...
+enum class ManagedMemoryType { Managed, Device, ManagedFine };
+#endif // HIP_VERSION_MAJOR
+#else
+enum class ManagedMemoryType { Managed, Device };
+#endif
+
+namespace config
+{
+
+class GtensorConfig
```
germasch, line 49: I guess the camel case comments goes for this, too, even though it's not public-facing, so it's less of an issue, but consistency is still nice.

germasch: I also don't see why this needs to be a class with getters/setters. A simple struct would do, right?

germasch (changes requested):
Still have some nitpicks ;)

include/gtensor/device_backend.h
bd4: The only other enum in current code base is enum Type in gslice.h, which seems to be camel case. I think there is something to be said for using a different convention for enums vs classes. It does use all caps for values, so at least that should be changed.

include/gtensor/device_backend.h
bd4: I think to make this work, I would need to use plain enum and not class enum. C++20 has using enum, but I don't think there is a way to use unqualified class enum before that?

bd4: I didn't figure out a way to make this work. The most obvious `managed_memory_type::GTENSOR_MANAGED_MEMORY_TYPE_DEFAULT` does not work.

germasch: Ok. I'm pretty sure it could be hacked (may require another macro and ##). But it doesn't seem worth it at this point.

germasch (approved):

include/gtensor/device_backend.h
bd4: found a surprisingly not too ugly solution for this

PR #199, opened 2022-08-18
refactor: add backend_common to fix lsp

bd4:
The backend includes only worked by being included in the right
spot inside device_backend.h, which is ugly and breaks tooling
in editors like nvim/vscode LSP. This just moves the required
parts of device_backend into a new backend_common header and
includes it directly from each backend header.


include/gtensor/backend_common.h
```diff
@@ -0,0 +1,300 @@
...
+  managed_fine
+};
+#endif // HIP_VERSION_MAJOR
+#else
+enum class managed_memory_type
+{
+  managed,
+  device
+};
+#endif
```
germasch, line 59: I actually forgot to make the comment on the last PR, and it's not something that belongs in this PR, but before I forget again:

I think we should generally try to make gtensor support having multiple backends available. That is, `GTENSOR_DEVICE_HIP` should not be exclusive to, e.g., also having defined `GTENSOR_DEVICE_CUDA` (which is not likely a common use case, but might still be useful to test both compile paths in one CI image. But there could, e.g., be some OpenMP backends in the future).

So in this particular case, I think one could just have all 4 memory types always in the enum, and then map them to what the hardware actually supports in the actual (backend-specific) implementation.

germasch (approved):

PR #201, opened 2022-08-18
Pr/bench stencil tmp

bd4:


germasch (commented):
Other than combining some unrelated changes in a single PR, looks good to me. I think doing multi-d stencils with arbitrary coefficient arrays is going to be more complicated still, but this definitely works for the benchmark.

bd4:
The unrelated changes is github being difficult when I try to stage a series of commits and seemingly not updating the diffs when I try to rebase or when the base changes, i.e. when the earlier PRs are merged. If you look at the diff view tab, it should show the real diff.

bd4:
hm I guess the diff tab also shows it. I guess I have to stage multiple PRs a different way. I'll rebase and force push

bd4:
Cleaned up, I believe is ready for merge now.

germasch (approved):
👍

PR #202, opened 2022-08-19
add typestr expression method for debugging

bd4:
Experimental way to add useful debug info, by calling `expr.typestr()`. This is far easier than template meta programming to do this separately, however it has the downside of requiring including string and sstream throughout the code, which I don't like.

### DIFF
--- a/tests/test_expression.cxx
+++ b/tests/test_expression.cxx
@@ -341,6 +341,24 @@ TEST(expression, shape_second)
   EXPECT_EQ(e.shape(), (S3{2, 3, 4}));
 }
 
+TEST(expression, typestr)
+{
+  gt::gtensor<double, 2> a = {{1, 2}, {3, 4}};
+  auto aview = a.view(gt::all, 0);
+
+  auto adfn = 5.0 + a;
+  auto avfn = 5 * aview;
+  auto asfn = 5 * a.to_kernel();
+  auto sina = gt::sin(a);
+
+  std::cout << "a     typestr " << a.typestr() << std::endl;
+  std::cout << "aview typestr " << aview.typestr() << std::endl;
+  std::cout << "adfn  typestr " << adfn.typestr() << std::endl;
+  std::cout << "avfn  typestr " << avfn.typestr() << std::endl;
+  std::cout << "asfn  typestr " << asfn.typestr() << std::endl;
+  std::cout << "sina  typestr " << sina.typestr() << std::endl;
+}
+
 template <typename S>
 void test_index_expression()
 {
--- a/tests/test_expression.cxx
+++ b/tests/test_expression.cxx
@@ -344,18 +344,36 @@ TEST(expression, shape_second)
 TEST(expression, typestr)
 {
   gt::gtensor<double, 2> a = {{1, 2}, {3, 4}};
+  // d2 -> d for owning data, 2 for 2d
+  // <double> for element type
+  // {2, 2} is shape
+  // {1, 2} is strides
+  EXPECT_EQ(a.typestr(), "d2<double>{2, 2}{1, 2}");
+  std::cout << "a     typestr " << a.typestr() << std::endl;
+
   auto aview = a.view(gt::all, 0);
+  // v1 is for 1d view
+  // (...) is the owning data object the view is over (above)
+  // {2}{1} are the strides of the view
+  EXPECT_EQ(aview.typestr(), "v1(d2<double>{2, 2}{1, 2}){2}{1}");
+  std::cout << "aview typestr " << aview.typestr() << std::endl;
 
   auto adfn = 5.0 + a;
-  auto avfn = 5 * aview;
-  auto asfn = 5 * a.to_kernel();
-  auto sina = gt::sin(a);
-
-  std::cout << "a     typestr " << a.typestr() << std::endl;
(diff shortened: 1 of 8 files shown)


bd4:
I think it is possible to do this with a free-standing function in a separate header not included by default. The complication is that private members would need to be exposed in some cases.

bd4:
Produces strings like this:
```
# owning (data) gtensor of 2 dimensions, element type double, shape (2,2) strides (1,2)
d2<double>{2, 2}{1, 2}
# 1d view of above, with shape (2) and strides (1)
v1(d2<double>{2, 2}{1, 2}){2}{1}
# function adding gscalar 5 as double to a 2d owning gtensor
fn:(5<double> + d2<double>{2, 2}{1, 2})
# function multiplying 5 as int with a 1d view of 2d owning gtensor
fn:(5<int> * v1(d2<double>{2, 2}{1, 2}){2}{1})
# function multiplying 5 as int with a 2d span
fn:(5<int> * s2<double>{2, 2}{1, 2})
# sin of 2d owning gtensor
fn:sin(d2<double>{2, 2}{1, 2})
```

benchmarks/gt-bm.h
```diff
@@ -26,6 +28,41 @@ struct gthelper<T, N, gt::space::managed>
...
+
+inline std::ostream& operator<<(std::ostream& o, string_view const& s)
+{
+  return o.write(s.data, s.size);
+}
+
+template <class T>
+constexpr string_view get_type_name()
+{
+  char const* p = __PRETTY_FUNCTION__;
```
germasch, line 47: I'm confused how this works, besides being a gcc extension only, I think. 

Isn't __PRETTY__FUNCTION__ alway sgoing to be `get_type_function` since that is the enclosing function? I though usually one has to use macros to be able to get the actual function name of interest (e.g., the way `assert` is a macro rather than an actual function).

include/gtensor/backend_common.h
```diff
@@ -293,6 +293,15 @@ struct selector
...
 
+// ======================================================================
+// prefetch interface
+
+template <typename T>
+void prefetch_device(T* p, size_type n);
+
+template <typename T>
+void prefetch_host(T* p, size_type n);
+
```
germasch, line 304: So I kinda don't like that as I'm looking at this, it's a mix of prefetch, the type printing and some benchmark changes, and if one ever has to bisect back to figure out where something broke, this could become an annoyance, too. I'd prefer if you broke this into a separate PRs.

germasch (commented):
This does seem like a useful addition. I don't think I mind it being added into the individual classes, otherwise one basically has to create a shadow hierarchy.

benchmarks/gt-bm.h
bd4: It works with LLVM and gcc, which covers the compilers we are using now. I don't understand exactly how it works, it's a bit of a hack.. but unfortunately there doesn't seem to be a good general truly standards compliant way to get demangled type names

bd4: I could add some comments to the hack. __PRETTY_FUNCTION__ looks like this:
```
const char* get_pretty_name() [with T = long unsigned int]
```
for `get_pretty_name<std::size_t>()`

include/gtensor/backend_common.h
bd4: That is separate commit and actually separate PR that has now been merged, I will rebase.

bd4:
The free function version of this does not require creating a shadow hiearachy, it can be done using function overloading. The main downside of having it separate is it complicates access to private members which is needed in many cases. One possible way to work around this, is to encapsulate the printer in a class and make it a friend of all the classes. But then potentially have to include all the extra headers anyway.

I'm going to benchmark compiling with different backends, to see if current version slows things down at all. If not, maybe this way is the simplest / best option.

I wonder if `typestr` is a good name for this - it does include shapes/sizes, which is not technically part of the type in the strict C++ since. I can't think of a better name though, `debugstr` seems confusing.

bd4:
Does not appear to affect compile times at all.

tests/test_expression.cxx
```diff
@@ -341,6 +341,24 @@ TEST(expression, shape_second)
...
+  auto sina = gt::sin(a);
+
+  std::cout << "a     typestr " << a.typestr() << std::endl;
+  std::cout << "aview typestr " << aview.typestr() << std::endl;
+  std::cout << "adfn  typestr " << adfn.typestr() << std::endl;
+  std::cout << "avfn  typestr " << avfn.typestr() << std::endl;
+  std::cout << "asfn  typestr " << asfn.typestr() << std::endl;
+  std::cout << "sina  typestr " << sina.typestr() << std::endl;
+}
+
```
germasch, line 361: I'm feeling a little lazy, so I'd prefer to be able to tell from the test what it's supposed to do ;) I'm pretty sure string `EXPECT_EQ()` will take strings ;)

germasch (changes requested):
This looks quite useful, but I'd like to see to the test actually test things, not just print them ;)

tests/test_expression.cxx
bd4: Not as lazy as when i wrote this test and saw it in the diff and didn't fix it ;). Pushed improved version that actually is a test, and has some comments on the mini-language used.

germasch (approved):
👍

PR #203, opened 2022-08-23
add cmake options

bd4:
add GTENSOR_CHECK_BOUNDS, GTENSOR_CHECK_ADDRESS, GTENSOR_SYNC_KERNELS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,8 @@ option(GTENSOR_PER_DIM_KERNELS
 
 option(GTENSOR_ALLOCATOR_CACHING "Enable naive caching allocators" ON)
 
+option(GTENSOR_BOUNDS_CHECK "Enable per access bounds checking" OFF)
+
 # HIP specific configuration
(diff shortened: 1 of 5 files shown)


bd4:
I wonder if check address should be on by default - I don't expect to gtensor_span objects to be construted in a loop, and it aids in early detection of some type of errors.

CMakeLists.txt
```diff
@@ -37,6 +37,10 @@ option(GTENSOR_PER_DIM_KERNELS
 
 option(GTENSOR_ALLOCATOR_CACHING "Enable naive caching allocators" ON)
 
+option(GTENSOR_BOUNDS_CHECK "Enable per access bounds checking" OFF)
+option(GTENSOR_ADDRESS_CHECK "Enable per access bounds checking" OFF)
```
germasch, line 41: A little copy&paste mistake here ;)

germasch (changes requested):
Looks great, other than one tiny bit! ;)

CMakeLists.txt
bd4: Good catch!

germasch (approved):

PR #204, opened 2022-08-29
add fortran interoperability

germasch:
This is basd on kokkos, essentially using their way of passing Fortran arrays (including non-contiguous ones). Adapted to create `gtensor_span` on the C++ side.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -130,11 +130,11 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: gtensor cmake
-      run: cmake -S . -B build-install -DGTENSOR_DEVICE=host -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/gtensor-install -DBUILD_TESTING=OFF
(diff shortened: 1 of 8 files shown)


bd4:
This looks great!

germasch:
The CI doesn't look like it's setup for Fortran (some images don't seem to have a Fortran compiler), but there's at least one case in there where the Fortran code gets compiled, so that's better than nothing.

There's also no real tests, just something that should probably better be moved into examples, and of course it wouldn't hurt
to add some actual tests, too.

germasch:
Not sure that I've got all the cmake subtleties right, but it looks like this works with gtensor as a subdir or via find_package if the consuming project is using cmake.

I haven't tried to make this work with regular make -- probably will have to add `-I...include/gtensor/fortran` to the Fortran command line, and link the right (`fgtensor`) library.

bd4 (approved):
LGTM. Regarding unit tests, I would prefer not adding pFUnit as a dependency to CI and general testing, I've had serious issues getting it to work on different machines. Eventually, maybe we can do something custom, even if it's just running some fortran exes and ensuring the return code is 0.

germasch:
I agree, I wouldn't want to pull in pfunit. I think it's actually pretty straight forward to convert my "example" into a test that checks rather than prints the passed array. I've just been too lazy ;)

PR #205, opened 2022-09-13
cmake: don't support old-style buildtree export

germasch:
I don't know anyone who's ever used this, and we already support regular installs as well as `add_subdirectory()`, which should suffice.

In particular this causes hassles integrating with other projects via `add_subdirectory()` that don't support buildtree export.


bd4 (approved):
LGTM, produces identical gtensor-targets.cmake so I guess it was redundant.

PR #206, opened 2022-09-14
sycl: use SYCL 2020 standard reductions

bd4:


germasch (approved):
Looks like something got simpler ;)

PR #207, opened 2022-09-15
add some support for running in host-only mode

germasch:
The main gtensor already mostly supports just using `<space::host>` but various `_device` functionality currently doesn't exist in this case. This starts making it available, where then, e.g. `adapt_device` just acts like `adapt<space::device>` and since `space::device == space::host`, it really just gives you a host gtensor_span, which makes it easier to write code that works with and with an actual gpu.

### DIFF
--- a/include/gtensor/backend_common.h
+++ b/include/gtensor/backend_common.h
@@ -26,6 +26,14 @@ namespace gt
 namespace backend
 {
 
+enum class memory_type
+{
+  host,
+  device,
+  managed,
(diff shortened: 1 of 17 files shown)


bd4:
Looks good

bd4 (approved):
LGTM, nice to codify the backend interface more in a class. Tests pass for me on HIP, CUDA, SYCL, and host with clang9, clang12, and gcc10.

PR #208, opened 2022-09-23
fortran: add gpu_api_m

germasch:
That is, basically, the Fortran side of what it takes to call cgtensor, the C API wrappers, though this really needs some cleaning up still. For now it's just exactly what used to live in reg_storage / GENE.


bd4 (changes requested):
I think gpu_api.cxx needs some refactoring - it contains quick hacks to get the direct CUDA code working on HIP, and later on some SYCL hacks were added. I'm fine merging as is and refactoring later, but something to consider. At the very least, some of this should be moved into C/C++ API as well, like the meminfo stuff.

germasch:
I agree that this will require more work, I think I already commented on that in a commit message. It's going to be a bit of pain, since some stuff (like prefetch) already exists in cgtensor but in a different form, so gene needs to be updated at the same time...

bd4 (approved):

PR #209, opened 2022-09-27
interface BLAS on the host

germasch:
This provides the BLAS interface C++ and C abstractions when compiled host-only. This required quite a bit of
boilerplate code, but oh well...

It only supports OpenBLAS right now, and it's missing `getrf_npvt_batched`, since I can't find anything like that in BLAS/LAPACK.


bd4:
Looks great! I am testing on a couple of machines this morning.

bd4 (approved):
I'm fine with this PR as is for first step (just openblas). Would be nice to support MKL eventually. Not sure about other blas implementations, I suppose there may be architectures that openblas does not support well, but not ones we use regularly currently.

PR #210, opened 2022-09-27
refactor stream_view

germasch:
Part of this is to not have every backend add a `gt::stream_view` type (ie., only one backend can be active). This also has the derived classes take more responsibility, rather than having the base class do a bunch of things via freestanding functions that are overloaded for each backend. There's the question whether the remaining bits really are worth CRTP...

More work t.b.d.


bd4:
Looks like a worth while refactor. Need to be careful about gtblas_*_stream because GENE currently uses them in the solver, arguably unnecessarily with how things ended up, but something to consider. If we merge this along with the solver refactor, then it wouldn't be an issue; otherwise we would need to remove the custom stream sets in GENE or have GENE use gt_create_stream or something, so it gets a stream_view pointer instead of a backend object. That would further force making sure no un-gtensor APIs are still being used that also get passed a stream object in that part of the code.

germasch:
Yeah, I know one needs to be careful with Fortran interfacing. I do think that right now, `sizeof(stream_view)` is always sizeof(void*) (or smaller), so it could always be stored as a `type(c_ptr)` on the Fortran side, so it should be doable to have a consistent interface, but it definitely needs to be checked before merging.

In general, I know we had the discussion before, but it does seem like it might be a good idea to make the BLAS handle_t into a full fledged class that has the various BLAS functions as member functions. In that case, the interfacing could be done via a pointer to that class, but I'm not even sure that necessarily helps with the stream_view interfacing.

germasch:
It's probably not a great idea to let stuff linger for a long time, but well, finally got back to it. I think this still has room for improvement, but well, I got it to pass the CI and I think it's progress.

I think the number one remaining concern is about the sycl queue references that have kinda different semantics from stream_view (at least, C++ thinks that references have different semantics), so I think I got that worked around for now, but it's not so pretty...

bd4 (approved):
I think this is a good improvement, thanks! Nice to have the stream_view be the consistent abstraction going into cblas/fortran. If we ever really need to allow a low level stream to be created in Fortran and set on gt-blas / gtensor, we can deal with it then, and I'm hoping we can avoid it.

Regarding the SYCL queue, according to spec they should have reference semantics, so passing them around by value should not result in any new low-level queues being created and the "copies" should be interchangeable. I had hard to debug problems when trying to do this in the past though, and am not eager to try again.

Lets merge this now, and we can iterate the the SYCL queue handling in due time.

PR #212, opened 2022-09-30
test: improved test predicates

bd4:
Adds overloaded GT_EXPECT_EQ(a, b), GT_EXPECT_NEAR(a, b), and GT_EXPECT_NEAR_MAXERR(a, b, max_err) macros that work with scalars, gtensor expressions, and combinations of the two (expressions must be in first position). The idea is for this to replace the confusingly named test_helpers.h header.

### DIFF
--- a/tests/test_gtest_predicates.cxx
+++ b/tests/test_gtest_predicates.cxx
@@ -1,0 +1,69 @@
+#include <string>
+
+#include <gtest/gtest.h>
+
+#include "gtensor/gtensor.h"
+
+#include "gtest_predicates.h"
+#include "test_debug.h"
+
+using namespace gt::placeholders;
+
+TEST(gtest_predicates, fp)
+{
+  double a = 1.0;
+  double b = 0.5;
+  GT_EXPECT_EQ(a, 2 * b);
+
+  float af = 1.0;
+  float bf = 0.5;
+  GT_EXPECT_EQ(af, 2 * bf);
+}
+
+TEST(gtest_predicates, complex)
+{
+  using D = gt::complex<double>;
+  D a = D(-1.0, 0.0);
+  D b = D(0.0, 1.0);
+  GT_EXPECT_EQ(a, b * b);
+
+  using F = gt::complex<float>;
+  F af = F(-1.0, 0.0);
+  F bf = F(0.0, 1.0);
+  GT_EXPECT_EQ(a, b * b);
+}
+
+TEST(gtest_predicates, complex_fp)
+{
+  using D = gt::complex<double>;
+  D im_d = D(0.0, -1.0);
+  GT_EXPECT_EQ(im_d * im_d, -1.0);
+
+  using F = gt::complex<float>;
+  F im_f = F(0.0, -1.0);
+  GT_EXPECT_EQ(im_f * im_f, -1.0);
+}
+
+TEST(gtest_predicates, gtensor_expr)
+{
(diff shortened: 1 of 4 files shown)


tests/test_gtest_predicates.cxx
```diff
@@ -0,0 +1,69 @@
...
+  D a = D(-1.0, 0.0);
+  D b = D(0.0, 1.0);
+  GT_EXPECT_EQ(a, b * b);
+
+  using F = gt::complex<float>;
+  F af = F(-1.0, 0.0);
+  F bf = F(0.0, 1.0);
+  GT_EXPECT_EQ(a, b * b);
+}
+
```
germasch, line 35: I guess a little more description of what this does, would have helped me to understand this a bit better, e.g. some output from when it fails. In particular, when using `GT_EXPECT_EQ` instead of `EXPECT_EQ` for real or complex numbers, I suppose it actually doesn't do anything different? Which as such might be reason to not use it for that, since if there are two ways of doing something, it makes one wonder as to whether the new way does something different?

I guess on the other side I can see that the idea might be to just use `GT_EXPECT_EQ` (and NEAR) always, since it also works on actual gtensors, so that one uses the same macro throughout? I might actually still think that `EXPECT_EQ` should work for gtensors in any case, and it's really only the `NEAR` version where we need something custom?

germasch (commented):
LGTM, but I guess I do have some questions still.

tests/test_gtest_predicates.cxx
bd4: The idea is to not have to think about which one to use, and the GT one supports gtensor arrays and prints a useful message (first 10 elements after mismatch). The _EQ just calls _NEAR with threshold of 0, so the messages are standardized for both. Another motivation, even for scalars, is to have one place to tweak failure, rather than having to do it all over when google test does not print anything very useful. I have also ran into issue where google test doesn't want to compare e.g. real and complex, which can be worked around by making a complex with 0 imaginary part, but doing it directly is nicer.

bd4: Some more details: googletest does provide `EXPECT_EQ(complex, real)`, but `EXPECT_NEAR(complex, real, max_err)` is not provided, and `max_err` is required. So the advantages of `GT_EXPECT_NEAR`:

1. provides complex, real version
2. provides default max error based on input types
3. supports (expr, expr) and (expr, scalar)

The `GT_EXPECT_EQ` helper is useful for expressions, the googletest version is fine for scalars, including (complex, real), and it's failure message is fine. The extension for GT_EXPECT_EQ on `(scalar, scalar)` is not necessary, but it does have the advantage of a uniform interface where one does not have to worry about what to use when.

I am in the process of improving the failure messages and having them be printed by the predicate test, so it's easier to see what the idea is.

bd4: The expr compare produces messages like this:

```
Expected near values (err=4 > 0.0001) at [1]:
  a
    Which is: { 1 2 3 }
  c
    Which is: { 1 -2 3 }
```

and

```
Expected near values (err=0.00099999999999988987 > 0.0001) at [0]:
  a
    Which is: { 1.0009999999999999 1.0000009999999999 1.0000000000010001 }
  1
    Which is: 1
```

germasch: Thanks, looks great ;)

germasch (approved):

PR #213, opened 2022-10-10
Pr/solver

bd4:
Adds high level solver interface:
```
#include "gt-solver/solver.h"
gt::blas::handle_t h;
gt::solver::SolverInvert<gt::complex<double>> s(h, N, NBATCHES, NRHS, matrix_batches);
s.solve(rhs, sol);
```

### DIFF
--- a/include/gt-solver/solver.h
+++ b/include/gt-solver/solver.h
@@ -15,6 +15,8 @@ template <typename T>
 class GpuSolverDense
 {
 public:
+  using value_type = T;
+
   GpuSolverDense(gt::blas::handle_t& h, int n, int nbatches, int nrhs,
                  T* const* matrix_batches);
 
@@ -41,12 +43,37 @@ extern template class GpuSolverDense<double>;
 extern template class GpuSolverDense<gt::complex<float>>;
 extern template class GpuSolverDense<gt::complex<double>>;
 
+template <typename T>
+class GpuSolverInvert : GpuSolverDense<T>
+{
+public:
+  using value_type = typename GpuSolverDense<T>::value_type;
+
+  GpuSolverInvert(gt::blas::handle_t& h, int n, int nbatches, int nrhs,
+                  T* const* matrix_batches);
+
+  virtual void prepare();
+
+  virtual void solve(T* rhs, T* result);
+
+protected:
+  gt::gtensor_device<T, 3> rhs_input_data_;
+  gt::gtensor_device<T*, 1> rhs_input_pointers_;
+};
+
+extern template class GpuSolverInvert<float>;
+extern template class GpuSolverInvert<double>;
+extern template class GpuSolverInvert<gt::complex<float>>;
+extern template class GpuSolverInvert<gt::complex<double>>;
+
 #ifdef GTENSOR_DEVICE_SYCL
 
 template <typename T>
 class GpuSolverDenseSYCL : public GpuSolverDense<T>
 {
 public:
+  using value_type = typename GpuSolverDense<T>::value_type;
+
   GpuSolverDenseSYCL(gt::blas::handle_t& h, int n, int nbatches, int nrhs,
                      T* const* matrix_batches);
 
--- a/include/gt-solver/solver.h
+++ b/include/gt-solver/solver.h
@@ -23,6 +23,39 @@ protected:
   virtual void prepare() = 0;
 };
 
+#ifdef GTENSOR_DEVICE_SYCL
+
+// use contiguous strided dense API for SYCL, it has been better
+// optimized
+
+template <typename T>
+class SolverDense : public Solver<T>
+{
+public:
+  using base_type = Solver<T>;
+  using typename base_type::value_type;
+
+  SolverDense(gt::blas::handle_t& h, int n, int nbatches, int nrhs,
+              T* const* matrix_batches);
+
+  virtual void solve(T* rhs, T* result);
+
+protected:
+  virtual void prepare();
+
+  gt::blas::handle_t& h_;
+  int n_;
+  int nbatches_;
+  int nrhs_;
+  gt::gtensor_device<T, 3> matrix_data_;
+  gt::gtensor_device<gt::blas::index_t, 2> pivot_data_;
+  gt::gtensor_device<T, 3> rhs_data_;
+  gt::blas::index_t scratch_count_;
+  gt::space::device_vector<T> scratch_;
+};
+
+#else // HIP and CUDA
+
 template <typename T>
 class SolverDense : public Solver<T>
 {
@@ -49,6 +82,8 @@ protected:
   gt::gtensor_device<T, 3> rhs_data_;
   gt::gtensor_device<T*, 1> rhs_pointers_;
 };
+
+#endif
 
 template <typename T>
 class SolverInvert : Solver<T>
@@ -78,36 +113,6 @@ protected:
   gt::gtensor_device<T, 3> rhs_input_data_;
   gt::gtensor_device<T*, 1> rhs_input_pointers_;
 };
-
-#ifdef GTENSOR_DEVICE_SYCL
-
-template <typename T>
-class SolverDenseSYCL : public Solver<T>
-{
-public:
-  using base_type = Solver<T>;
-  using typename base_type::value_type;
-
-  SolverDenseSYCL(gt::blas::handle_t& h, int n, int nbatches, int nrhs,
-                  T* const* matrix_batches);
-
-  virtual void solve(T* rhs, T* result);
-
-protected:
-  virtual void prepare();
-
-  gt::blas::handle_t& h_;
-  int n_;
-  int nbatches_;
-  int nrhs_;
-  gt::gtensor_device<T, 3> matrix_data_;
-  gt::gtensor_device<gt::blas::index_t, 2> pivot_data_;
-  gt::gtensor_device<T, 3> rhs_data_;
-  gt::blas::index_t scratch_count_;
-  gt::space::device_vector<T> scratch_;
-};
-
-#endif
 
 } // namespace solver
 
--- a/include/gt-solver/solver.h
+++ b/include/gt-solver/solver.h
@@ -1,0 +1,46 @@
+#ifndef GTENSOR_SOLVE_H
+#define GTENSOR_SOLVE_H
+
+#include "gtensor/gtensor.h"
+
+#include "gt-blas/blas.h"
+
+namespace gt
+{
+
+namespace solver
+{
+
+template <typename T>
+class GpuSolverDense
+{
+public:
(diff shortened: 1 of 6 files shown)


src/gtsolver.cxx
```diff
@@ -0,0 +1,191 @@
...
+{
+  gt::gtensor<T*, 1> h_matrix_pointers(gt::shape(nbatches));
+  gt::gtensor<T*, 1> h_rhs_pointers(gt::shape(nbatches));
+  auto pdata = matrix_data_.data();
+  decltype(pdata) pmatrix;
+  T* prhs_data = gt::raw_pointer_cast(rhs_data_.data());
+  gt::blas::index_t* ppivot_data = gt::raw_pointer_cast(pivot_data_.data());
+  // copy in to contiguous device memory
+  for (int i = 0; i < nbatches; i++) {
+    pmatrix = pdata + n * n * i;
```
germasch, line 34: You could just use `auto pmatrix = pdata + ...`, right?

src/gtsolver.cxx
```diff
@@ -0,0 +1,191 @@
...
+template <typename T>
+void GpuSolverDense<T>::prepare()
+{
+  gt::blas::getrf_batched<T>(h_, n_,
+                             gt::raw_pointer_cast(matrix_pointers_.data()), n_,
+                             gt::raw_pointer_cast(pivot_data_.data()),
+                             gt::raw_pointer_cast(info_.data()), nbatches_);
+  prepared_ = true;
+}
+
```
germasch, line 52: Is there a reason that `prepare()` can't just be called in the constructor? That way, the object wouldn't have a "initialized but not yet usable for solve" kind of state, ie., it'd always be ready to solve and one wouldn't need `prepared_`, either, I'd think?

src/gtsolver.cxx
```diff
@@ -0,0 +1,191 @@
...
+  }
+  gt::copy(h_rhs_input_pointers, this->rhs_input_pointers_);
+}
+
+template <typename T>
+void GpuSolverInvert<T>::prepare()
+{
+  // factor into matrix_data_
+  GpuSolverDense<T>::prepare();
+
```
germasch, line 93: Maybe that's the reason for the separate `prepare`, but I'd still prefer to have that done internally / with private methods. 

The other somewhat related question here is, is it anticipated that the classes are going to be used through a pointer-to-base class, or is the inheritance just because they share some common code? I think in the former case, it might make sense to have an abstract base class, while otherwise one shouldn't actually need virtual functions?

germasch (commented):
Looks nice, but I did come up with a couple of questions ;)

include/gt-solver/solver.h
```diff
@@ -0,0 +1,100 @@
...
+
+protected:
+  gt::blas::index_t scratch_count_;
+  gt::space::device_vector<T> scratch_;
+};
+
+extern template class GpuSolverDenseSYCL<float>;
+extern template class GpuSolverDenseSYCL<double>;
+extern template class GpuSolverDenseSYCL<gt::complex<float>>;
+extern template class GpuSolverDenseSYCL<gt::complex<double>>;
```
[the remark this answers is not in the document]
bd4: Just the definition in the .cxx seems to be enough, I'll remove the extern declarations.

src/gtsolver.cxx
bd4: Two reasons, one is to be able to benchmark the preparation step, and the other reason is code re-use and sharing. The latter could probably still be accomplished by making prepare a protected method called at the end of the constructor, but there is less flexibility. The idea is the constructor does the allocations and setting up pointer arrays, and prepare does computation, e.g. factor or inversion.

I agree it's ugly having to explicitly call prepare. I'll see if I can get the code re-use I want if it's called at the end of the constructor. Also there isn't that much code, so some duplication is not the worst.

src/gtsolver.cxx
bd4: Potentially both - they all have the same interface and several of the solver implementations can share code. I think having an abstract base class defining the interface is worth it, and passing a pointer to base solver around could be useful.

I do wonder if it's worth trying to share code - figuring out exactly what arrays are used for which and how to share them most effectively is confusing, and while it does make for less duplication, it makes the code harder to follow, and it's not that much code in the end. So just having an abstract base class to define the interface and allow passing implementations around, is maybe a better design.

These are the likely set of initial implementations:
1) standard dense solve, getrf / getrs
2) invert solve - getrf + getri / gemm
3) alternate dense solve for oneMKL - use strides APIs, getrf / getrs, don't need any of the pointer arrays, but needs additional temporaries based for scratch
4) sparse solve - dense getrf nopivot + convert to single CSR / sparse trsv x2

I don't think it's worth adding the naive band solver, it out-performes dense in some cases but on those platforms the invert solve is pretty much always better, gemm_batch seems to have been optimized everywhere

bd4: rather than sharing code via inheritance, they could share code via helper methods, .e.g `init_device_pointer_array(d_pointers, d_data, data_stride, nbatches)`. Most of the meat is in gt-blas calls anyway, the sharing is just tedious array setup stuff.

bd4:
Does not yet have a sparse implementation, but that will be more complicated because of the lack of a existing gt-sparse wrappers and the larger differences among sparse APIs. Seems sensible to merge this first as a starting point.

src/gtsolver.cxx
```diff
@@ -0,0 +1,216 @@
...
+    matrix_data_(gt::shape(n, n, nbatches)),
+    pivot_data_(gt::shape(n, nbatches)),
+    rhs_data_(gt::shape(n, nrhs, nbatches)),
+    scratch_count_(gt::blas::getrs_strided_batched_scratchpad_size<T>(
+      h, n, n, n, nrhs, nbatches)),
+    scratch_(scratch_count_)
+{
+  detail::copy_batch_data(matrix_batches, matrix_data_);
+  prepare();
+}
```
germasch, line 63: I think one question here is whether to not just do the preparation here directly, though I don't think I have strong feelings about that (it's not particularly complicated, but kinda ugly-ish with those blas interfaces, so having separate does help a little in terms of keeping things compartmentalized.

However, for what I've seen so far, I don't think `prepare` needs to be virtual, and in fact since you're calling it from the constructor, it will not truly be virtual in any case. (I'd argue it's also not required for every subclass of `Solver` to necessarily have a `prepare` function, it might not really be needed in some cases, e.g., when one has an iterative solver that doesn't need any kind of factorization).

src/gtsolver.cxx
```diff
@@ -0,0 +1,216 @@
...
+{
+  auto prep_scratch_count =
+    gt::blas::getrf_strided_batched_scratchpad_size<T>(h_, n_, n_, nbatches_);
+  gt::space::device_vector<T> prep_scratch(prep_scratch_count);
+
+  gt::blas::getrf_strided_batched<T>(
+    h_, n_, gt::raw_pointer_cast(matrix_data_.data()), n_,
+    gt::raw_pointer_cast(pivot_data_.data()), nbatches_,
+    gt::raw_pointer_cast(prep_scratch.data()), prep_scratch_count);
+  // Note: synchronize so it's safe to garbage collect scratch
```
germasch, line 76: I don't think C++ does garbage collection ;) (I do agree that the sync is needed).

tests/test_solver.cxx
```diff
@@ -0,0 +1,112 @@
...
+  }
+  h_Aptr(0) = gt::raw_pointer_cast(h_A.data());
+
+  gt::blas::handle_t h;
+
+  Solver solver(h, N, batch_size, NRHS, gt::raw_pointer_cast(h_Aptr.data()));
+
+  gt::copy(h_B, d_B);
+  solver.solve(gt::raw_pointer_cast(d_B.data()),
+               gt::raw_pointer_cast(d_C.data()));
```
germasch, line 57: off-topic, but all those pointer_cast I've been seeing here and elsewhere lately really makes me think we should seriously consider just using raw pointers throughout.

tests/test_solver.cxx
```diff
@@ -0,0 +1,112 @@
...
+  gt::copy(d_C, h_C);
+
+  gt::gtensor<T, 1> h_C_expected(gt::shape(N));
+
+  h_C_expected(0) = 2.5;
+  h_C_expected(1) = 4.0;
+  h_C_expected(2) = 4.5;
+  h_C_expected(3) = 4.0;
+  h_C_expected(4) = 2.5;
```
germasch, line 66: I'd still prefer to write this more compactly like `h_C_expected{2.5, 4.0, 4.5, 4.0, 2.5}`, though not sure whether in practice there might be issues with conversion to complex, etc.

germasch (changes requested):
Looks nice -- the only thing I'd still like to see is to have `prepare()` not be virtual (unless I missed something).

In terms of the sparse stuff, I agree with holding off. I think it may not be advisable to try to create a blas-like abstraction layer there, anyway, but basically have a `SolverCuSparse` that directly calls cusparse as needed. (Though cuda and hip may be so similar that it may make sense to have one class for both of them).

The other thing I just noticed as I typed the above is the camel case used here, while I believe all of the rest of gtensor does follow the (rather old) lower case + underscore convention. It'd be kinda nicer to stay consistent with that.

src/gtsolver.cxx
bd4: I removed prepare and just moved it into the ctors.

src/gtsolver.cxx
bd4: I moved it in to the constructors, I think it wasn't serving any real purpose for re-use, and for benchmarking I can always bench the constructor. I don't love that the constructor ends up being so expensive, but still seems like the best design for this, rather than having a extra step.

src/gtsolver.cxx
bd4: Even though I haven't used Java in years, it has definitely left it's mark when I do OO programming :P. Fixed in latest commit.

tests/test_solver.cxx
bd4: It is used for `gt::copy_n` to determine which enum to pass to low level cudaMemcpy or similar. Another possibility would be to make the gt-(blas,solver,etc) APIs able to take non-raw pointers, and just remove the layer at the end before calling the low level API. I think raw_pointer_cast just returns arg if already raw, or could be made to.

tests/test_solver.cxx
bd4: ya I've been meaning to try to debug that, it is annoying having to be so verbose.

tests/test_solver.cxx
germasch: Good point, there's definitely stuff that'd break if used untagged pointers. So that might well not be a win, but definitely not easy. So I'm giving up on the thought for now... ;)

germasch (approved):
👍

PR #214, opened 2022-10-25
fortran: use iso_fortran_binding.h for Fortran interop

germasch:
instead of pulling in Kokkos' interop layer

As long as it works, this is clearly the nicer way of doing things, and it does look like this is supported with the current generation of Fortran compilers. (tested with gfortran and Intel, anyway)


bd4 (approved):
Works for me with both ifort and ifx with dpcpp as the linker, if I make some changes to the CMakeLists.txt to get dpcpp to link the fortran main properly. I'm fine just making that a separate PR, as this does not break any existing functionality and is still passing CI.

PR #219, opened 2022-11-09
reductions: add specific thrust headers

bd4:
As of rocthrust 5.3.0, the reduce.h header no longer include max_element and transform_reduce. The good news is that the fine grained headers appear to have existed in both cuda nad rocm thrust for a while, so adding them fixes 5.3.0 without breaking older releases.


germasch (approved):
👍

PR #225, opened 2022-11-15
Pr/sycl ext complex

bd4:
ALCF is working to standardize `sycl::complex`, and has a library implementation available with `sycl::ext::cplx::complex`. This work nicely in gtensor, except when passing pointers around, see https://github.com/argonne-lcf/SyclCPLX/issues/19. This PR includes a lightly modified version of the library header from SyclCPLX, along with some template meta programming ugliness to convert to std::complex in gt-blas, gt-fft, and gt-solver. Once this is standardized and oneMKL is updated, we should be able to remove the hacks. In the meantime, this allows us to contribute towards the standardization process by providing a real world test case.

### DIFF
--- a/.github/workflows/sycl.yml
+++ b/.github/workflows/sycl.yml
@@ -10,15 +10,15 @@ on:
 jobs:
   test-sycl:
     runs-on: ubuntu-latest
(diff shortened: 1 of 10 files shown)


germasch (approved):
As usual with sycl stuff, I took a look but not really in depth ;)

PR #228, opened 2022-11-29
ub: fix zero length array in zero dim shapes

bd4:


germasch:
So I guess I probably need to look at the code to refresh my memory. I agree that things are broken in that zero-length arrays are not allowed. `std::array` with 0 length is allowed, though, so I guess that's a good place to go from. I'm not even sure why `gt::sarray` exists -- I suppose it should behave like `std::array` but be device accessible, ie., has `GT_INLINE` to allow functions to be called in device code. So I guess that's kinda similar, if annoying, to the `gt::complex` situation. I wonder, though, whether the nvcc option to make constexpr stuff usable on the device might be enough so that we could use `std::array` directly, that'd be an opportunity to simplify things.

In any case, though, this mostly looks good, except that I don't understand why you still added a length-1 data array to it -- any attempt to actually use `operator[]` on a 0-length array is illegal, right?

bd4:
> In any case, though, this mostly looks good, except that I don't understand why you still added a length-1 data array to it -- any attempt to actually use `operator[]` on a 0-length array is illegal, right?

The compiler wants it to be there, because the `to_string` method calls, but it should never be called. I could try to overload `to_string` instead, which may require the struct detail TMP trick. I am not sure what is least ugly here. It is wasteful having that data, adding to the size of each gscalar object. I can experiment a bit more with it. Might be interesting to look at how std::array handles the zero case too.

germasch (approved):
LGTM

PR #229, opened 2022-12-09
sarray: fix type used for index in operator[]

germasch:
A lot of the time, this has been used for shapes etc, where T is an integer type, so it probably wasn't harmful, but still, it's clearly buggy to use `T` for the index type.


bd4 (approved):
LGTM, good catch

PR #230, opened 2022-12-14
add sparse matrix type

bd4:
Adds sparse matrix types and fixes ROCm CI. In preparation for adding a sparse solver to gt-solver.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -33,7 +33,7 @@ BraceWrapping:
   BeforeCatch:     false
   BeforeElse:      false
   IndentBraces:    false
-  SplitEmptyFunction: true
+  SplitEmptyFunction: false
   SplitEmptyRecord: false
   SplitEmptyNamespace: true
 BreakBeforeBinaryOperators: None
--- a/tests/test_sparse.cxx
+++ b/tests/test_sparse.cxx
@@ -8,27 +8,6 @@
 #include "gtest_predicates.h"
 #include "test_debug.h"
 
-namespace detail
-{
-
-template <typename T>
-GT_INLINE T norm(gt::complex<T> a)
-{
-  return gt::norm(a);
-}
-
-GT_INLINE double norm(double a)
-{
-  return gt::abs(a);
-}
-
-GT_INLINE float norm(float a)
-{
-  return gt::abs(a);
-}
-
-} // namespace detail
-
 template <typename T, typename S>
 void test_csr_matrix_batched()
 {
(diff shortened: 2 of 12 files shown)


bd4:
I ended up having to fix some ROCm CI issues while working on this - I think it makes sense to merge, and do the actual solver parts as a separate PR.

bd4:
Even though this has no external dependencies currently, I wonder if it should be a sub-lib, e.g. `include/gt-sparse`. The conversion from dense to sparse could be improved by calculating the row start /end pointers (row_ptr array) at the same time as calculating nnz. Then the conversion can be parallelized on row, not just on batch. This is tricky, because the row_ptr calculation is a custom kernel, and the nnz totally is still a reduction. One way to handle is to use CUB on cuda, rocPRIM for rocm, and SYCL resucitons in sycl, to try to do it all in the same kernel. The other way would be to calculate the row_ptr in a custom kernel, parallelized on row (which takes into account batch), then use a generic reduction as a separate kernel, or do it on host since nrows is going to be relatively small.

It may workout that using vendor specific libraries doesn't improve performance or readability at all or enough to justify the complexity. Just makes me wonder if it should be separate, just in case.

.clang-format
```diff
@@ -33,7 +33,7 @@ BraceWrapping:
   BeforeCatch:     false
   BeforeElse:      false
   IndentBraces:    false
-  SplitEmptyFunction: true
+  SplitEmptyFunction: false
```
germasch, line 36: Hmmh, I guess generally I prefer not to change formatting options unless there's a good reason. But I don't see any changes in the diff here because of it, and I guess I'd agree hat I like `{}` better -- and that's mostly already the case at least for inline constructors, which I suppose may have their separate clang-format option.

tests/test_sparse.cxx
```diff
@@ -0,0 +1,231 @@
...
+GT_INLINE T norm(gt::complex<T> a)
+{
+  return gt::norm(a);
+}
+
+GT_INLINE double norm(double a)
+{
+  return gt::abs(a);
+}
+
```
germasch, line 24: What (if anything) does `gt::norm` currently return for a non-complex number? This looks like an opportunity to have `gt::norm` behave like `std::norm`. Noteworthy: `std::norm` returns **the square** of the absolute value.

germasch (approved):
LGTM, with some minor questions. (The `gt::norm` thing certainly could be done separately.)

.clang-format
bd4: The thing is, the code is formatted as if this was false already - clang-format-9 appears to not have supported this option and behaved as if it was false already. When using a newer clang-format version, this is fixed, and it tries to reformat everything that passes CI with clang-format-9 and breaks it. We could update to a newer verison in CI and have everyone use a newer version, and change all the files to use the separate line brackets, but changing the config to reflect the reality seems simpler.

tests/test_sparse.cxx
bd4: `std::norm` is not defined directly for double/float, but I guess the implicit conversion kicks in and it constructs a complex with real part matching the passed double or float. For some reason this does not work with some GPU backends in device code. Maybe a better solution would be to tweak the defs in complex_ops.h to make it work more like std::norm, even if that means defining explicit real versions.

.clang-format
germasch: I like it better with `false`, anyway, I just didn't understand the background -- thanks!

tests/test_sparse.cxx
germasch: According to this, it is defined: https://en.cppreference.com/w/cpp/numeric/complex/norm.

It's true that std::whatever often doesn't work with `gt::complex`, well, because the latter often isn't `std::complex`. But I think we agree, the best solution would be to have `gt::norm` support float/double, too.

PR #232, opened 2022-12-24
Pr/sparse solver2

bd4:
Adds sparse implementations for all backends to gt-solver. The SYCL backend does not support complex yet, because the underlying MKL API does not (it is planned for future).


bd4:
Note that the cuda backend uses the older csrsm solver API. The new generic solver API, `cusparseSpSV` (single rhs) and `cusparseSpSM` (multiple rhs), were added in different 11.X releases and the old csrsm API was deprecated at some point. This is still too new to require a release with the new API.

The new API is different enough, adding it as an separate alternate cuda backend is probably the simplest way to implement it. Then we could conditionally use the new API based on CUDART_VERSION, that way when the deprecated API is removed everything will continue to work.

bd4:
Added new CUDA API backend.

bd4:
Initial testing with new cusparse solve API in CUDA 12, is it's good for performance and bad for memory usage. Which unfortunately removes it's purpose, because invert solve and even dense are almost always faster and it no longer uses less mem, at least in my tests so far.

CMakeLists.txt
```diff
@@ -185,6 +185,11 @@ if ("hip" IN_LIST GTENSOR_BUILD_DEVICES)
                         "${ROCM_PATH}/lib/librocsolver.so")
   target_include_directories(rocsolver INTERFACE "${ROCM_PATH}/include")
 
+  add_library(rocsparse INTERFACE IMPORTED)
+  target_link_libraries(rocsparse INTERFACE
+                        "${ROCM_PATH}/lib/librocsparse.so")
+  target_include_directories(rocsparse INTERFACE "${ROCM_PATH}/include")
+
```
germasch, line 192: Side comment: I think at some point we should try to use the proper cmake support for hip/rocm, which I'd hope should actually be usable by now.

README.md
```diff
@@ -455,3 +455,67 @@ native stream object is a `sycl::queue`.
...
 particular, the stream objects will become templated on space type.
+
+# Library Wrapper Extensions
+
+## gt-blas
+
+Provides wrappers around commonly used blas routines. Requires cuBLAS, rocblas,
+or oneMKL, depending on the GPU backend. Interface is mostly C style taking
+raw pointers, for easy interoperability with Fortran, with a few higher level
+gtensor specific helpers.
```
germasch, line 466: 👍 for adding some documentation ;)

germasch (approved):
I can't claim that I've really gone through this carefully, but since it's tested and almost all new stuff, I'm pretty sure it won't break things ;)

CMakeLists.txt
bd4: Agreed, I created an issue for it: https://github.com/wdmapp/gtensor/issues/233

PR #235, opened 2023-01-05
add gt::host_mirror

germasch:
All it does is allocate a gtensor on the host with matching size.
    
It does not copy either way, that needs to be done explicitly -- that's one way to allow the user to specify whether they're reading from, or writing to the mirrored gtensor.

### DIFF
--- a/tests/test_gtensor.cxx
+++ b/tests/test_gtensor.cxx
@@ -786,3 +786,38 @@ TEST(gtensor_kernel, kernel_lambda_call)
 }
 
 #endif
+
+template <typename S>
+class gtensor_space : public ::testing::Test
+{};
+
+using gtensor_space_types = ::testing::Types<
+#ifdef GTENSOR_HAVE_DEVICE
+  gt::space::device,
+#endif
+  gt::space::host>;
+
+TYPED_TEST_SUITE(gtensor_space, gtensor_space_types);
+
+// host_mirror should basically be a no-op when compiling host only (space_type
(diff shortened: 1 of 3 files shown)


tests/test_gtensor.cxx
```diff
@@ -769,7 +769,10 @@ void lambda_test(const gt::gtensor_device<double, 1>& a,
   auto k_a = a.to_kernel();
   auto k_b = b.to_kernel();
 
-  auto lf = GT_LAMBDA(int i) mutable { k_b(i) = k_a(i); };
+  auto lf = GT_LAMBDA(int i) mutable
+  {
+    k_b(i) = k_a(i);
+  };
```
bd4, line 775: This might be the cause of the clang format failure in CI? Maybe we should update to a newer version of clang-format, the oldest available in ubuntu 22.04 main repos is 11, and the default is 14; something in that range seems reasonable. This also likely means that we are missing an option in .clang-format, and a default changed in newer versions or was fixed. Also, this does not need to be mutable anymore.

bd4:
I like this, I guess a parameterized `gt::empty_like<gt::space::host>` would always allocate a new object, whereas host_mirror returns the original if already on host?

bd4 (approved):
looks good once format is fixed for CI. Edit: also needs a rebase.

tests/test_gtensor.cxx
bd4: I created an issue for this #236  to track separately, looks like you fixed the format issue for this pr.

PR #237, opened 2023-01-08
improve gt::copy

germasch:
This at least attempts to handle all kinds of different cases how gt::copy may be used. It fixes at least one existing bug
(copies involving non-contiguous gtensor_spans), but now also handles having expressions (rather than just gtensor/gtensor_span) as source and/or destination.

There might still be some cases that I missed, hopefully the kind that will lead to compile time errors, but the cases I handle are all covered by tests, so should be okay.

### DIFF
--- a/include/gtensor/backend_common.h
+++ b/include/gtensor/backend_common.h
@@ -248,6 +248,13 @@ GT_INLINE auto device_pointer_cast(T* p)
   return pointer(p);
 }
 
+template <typename S = gt::space::device, typename T>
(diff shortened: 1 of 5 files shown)


bd4 (approved):
Looks good! Not needed as part of this PR, but I wonder if it's worth having an adapt that checks the pointer with is_device_accessible and selects the appropriate space. This would be useful in the context of gt-solver, but of course it could also just call is_device_accessible itself.

Issue #236, opened 2023-01-05
ci: update clang-format version


bd4:
Ubuntu 22.04 has been out for 9 months, and it does not have clang-format-9 in main repo - the default version is 14, and the oldest is 11. Perhaps we should update CI to something in between, like version 12? @germasch what do you think?

bd4:
12 seems like the right version - it's the highest version available in old lts, 20.04, so anyone using an old docker image can still get it easily.

germasch:
Well, it'd be nice first of all if different versions of clang-format would do formatting consistently, which I guess is not always the case :(

I haven't looked into any of this in detail -- if I work on crusher, I end up with a mis-formatted bit, and I guess that's from
Cray clang-format version 14.0.2.

The ideal thing would be to find some way to change the config such that it doesn't matter whether it's clang-format 14, or 9, or somewhere in between. Not sure that's possible. This is the diff I've been getting:

```
--- ./tests/test_gtensor.cxx	(original)
+++ ./tests/test_gtensor.cxx	(reformatted)
@@ -775,[10](https://github.com/wdmapp/gtensor/actions/runs/3865193086/jobs/6588531541#step:4:11) +775,7 @@
   auto k_a = a.to_kernel();
   auto k_b = b.to_kernel();
 
-  auto lf = GT_LAMBDA(int i) mutable
-  {
-    k_b(i) = k_a(i);
-  };
+  auto lf = GT_LAMBDA(int i) mutable { k_b(i) = k_a(i); };
 
   gtLaunchKernel(kernel_test_lambda, 1, 3, 0, 0, b.shape(), lf);
 }```

I guess otherwise I'll agree that 12 seems like a reasonable choice, but I actually don't know if it behaves more like 9, or like Cray-14, or yet differently..

bd4:
The problem appears to be that newer versions of clang-format don't recognize `GT_LAMBDA(...) ...` as a lambda. Even setting  `AllowShortLambdasOnASingleLine: All` in the config does not change and it wants to expand it. Not obvious to me how to work around this.

bd4:
clang 14 is where this changed, and it is still that way in 15, so does not seem like it will change.

[...]

PR #238, opened 2023-01-08
format: single line all functions, dual version ci

bd4:
Starting in v14, short GT_LAMBDA is broken into multiple lines, which creates an inconsistency depending on version used. This change makes it consistent and adds ci checks for both 9 and 14.


bd4:
FWIW, using multiple clang format versions is basically free, it's an extra 2 seconds or even less if it parallelizes. So we could do every version supported by the github action, I think 9-15.

bd4:
fixes #236

germasch (approved):
Great, thanks!

PR #239, opened 2023-01-16
update sycl_ext_complex

bd4:
Note that for some reason the gtensor specified AssignN kernel names no longer work with the additional template parameter on sycl ext complex type; the name is missing the void parameter on some names in the kernel and it appears to be a phase where the default parameter can't be inserted, and it fails to compile.


bd4:
See https://github.com/argonne-lcf/SyclCPLX/pull/25

germasch (approved):
Looks fine to me.

PR #240, opened 2023-01-18
sarray: silence 0 len no return warning

bd4:
Note that __builtin_unreacheable works on device starting with cuda 11.3


germasch (approved):
👍

PR #242, opened 2023-01-19
build: replace use of GTENSOR_DEFAULT_DEVICE with GTENSOR_DEVICE

germasch:
Currently, GTENSOR_DEFAULT_DEVICE is not set, leading to a broken gtensor-config.cmake. I suppose one could set it on the cmake command line, but it seems overly complicated. I'm not sure that this is the best final solution, but it should be an improvement, and hopefully fix the CI problem I've been having with my gtensor-using code.


bd4 (approved):
LGTM. I think this has been broken since it was introduced :P.

PR #243, opened 2023-01-26
AMD: Take ROCM_PATH from ENV if possible and adjust the offload flag …

gmerlo:
…(avid deprecation warning)


bd4 (approved):
Looks great, thanks! I tested and offload-arch already works in ROCm 4.5.0, so I think is safe to make the change.

PR #245, opened 2023-02-03
Pr/bbfft

bd4:


uphoffc:
Hi Bryce,

great that you're picking up the double-batched FFT library. I have a few remarks:

1) You call e.wait() in the forward but not in the backward transform. Is that intended?
2) For the stride calculation you have
```c++
    std::array<std::size_t, bbfft::max_tensor_dim> rstride = {
      istrideu, istrideu * shape[0], istrideu * shape[1] * shape[0]};
```
But currently rstride[0] != 1 is not supported (and similar for cstride). Instead you should use
```c++
    std::array<std::size_t, bbfft::max_tensor_dim> rstride = {1u, istrideu, istrideu * shape[0]};
```
Such that, e.g., for the MxN1xK tensor you get
x[m * rstride[0] + n1 * rstride[1] + k * rstride[2]] = x[m + n1 * istrideu + k * istrideu * shape[0]]
Because you always set M=1 you effectively get
x[n1 * istrideu + k * istrideu * shape[0]]
analogous to FFTW-like interfaces.
3) The valid_ bool is not necessary. The only member of the plan class is a shared ptr. The default-constructed plan initializes the shared_ptr with nullptr and the class overrides operator bool. Hence, you can just use if (plan) to check whether the plan is valid. Moreover, if the plan is moved the "other" plan is set invalid by the default move constructor of shared_ptr. Here's a little reproducer of the behaviour of plan: https://godbolt.org/z/sohEM5aTr

Best regards,
Carsten

bd4:
@uphoffc thanks for the review! The gtensor tests using non-unit strides actually pass, so it seems that non-unit stride works in some cases, maybe for the $M=1$ case? I can certainly raise a runtime exception when non-unit stride is passed, working by luck with a particular release is not good. Do you plan on adding non-unit stride support in the future?

uphoffc:
Do you have any application in mind where one would need bbstride[0] != 1?
I have tested double-batching in GENE and there it is not necessary.

bd4:
Thanks for the detailed explanation! No application in mind for bbstride[0] != 1. Perhaps the simplest thing for now, is to throw an exception if non-standard layout is requested, i.e. if in the gt-fft interface, istride/ostride or idist/odist is not the standard, and call the bbfft function to set standard layout on the configs. And modify tests to expect exceptions for non-standard layout tests, ifdef on bbfft backend being enabled. This certainly meets the needs of using bbfft in GENE. It might be interesting to expose the double batching in the gt-fft interface somehow as well, currently it just uses bbfft similar to how it uses oneMKL in the default SYCL backend, no additions to the interface.

bd4:
@uphoffc any plans to support out of place for 2/3d fft? 2d FFT is used in GENE by the local code. I can work around this fairly easily by doing a copy from input to output and doing inplace in output, just wondering.

bd4:
> @uphoffc any plans to support out of place for 2/3d fft? 2d FFT is used in GENE by the local code. I can work around this fairly easily by doing a copy from input to output and doing inplace in output, just wondering.

Not so simple actually because of complex data is slightly bigger from the +1 in N/2 + 1 for the inner most dim, so inverse transform needs to use a temporary or plan lifetime buffer. Gets more complicated with non standard layout, which makes less sense for non standard layout where the simplest thing would be to have the temporary follow whatever the non-standard layout, but that wastes space.

uphoffc:
> @bd4 any plans to support out of place for 2/3d fft? 2d FFT is used in GENE by the local code. I can work around this fairly easily by doing a copy from input to output and doing inplace in output, just wondering.

I just added support for out-of-place 2d/3d fft in version 0.3.6. Using that version the 2d and 3d tests run through (although I had to slightly increase the error tolerance from 20 * max_err to 22 * max_err).

There is a limitation for 2d / 3d, though: Any strides that deviate from set_strides_default(true) or set_strides_default(false) won't work, i.e. only the default data layout is supported. I need to wrap my head around how to implement that efficiently. (An easy solution would be to copy-in the non-default layout to the default first, run an in-place FFT, and then copy-out. But these are 2 extra copies, so I'd like to think about a better solution first when I find some time.)

> Perhaps the simplest thing for now, is to throw an exception if non-standard layout is requested, i.e. if in the gt-fft interface, istride/ostride or idist/odist is not the standard, and call the bbfft function to set standard layout on the configs.

So for 1d non-standard istride/ostride and idist/odist are ok, so no need to throw an exception. For 2d/3d, yes, see comment above.

> It might be interesting to expose the double batching in the gt-fft interface somehow as well, currently it just uses bbfft similar to how it uses oneMKL in the default SYCL backend, no additions to the interface.

One could make the interface such that the shape / stride of a gtensor is passed and one specifies the modes over which the FFT is to be taken. It's not so easy, though, as one needs to figure out how to map all the possible cases to the FFT back-end.
I'll write you a separate message about what I actually did in GENE.

bd4:
HIP CI is failing because repo.radeon.com is having problems

bd4:
Cleaned up commits with an interactive rebase, I believe it's ready to merge now.

germasch (approved):
LGTM.

PR #246, opened 2023-02-10
Improved generic bandsolver for SYCL

uphoffc:
An optimized generic bandsolver written in SYCL. It would be better to not organize diverging code paths via #ifdefs so I am open to suggestions on how you would like to handle these.

### DIFF
--- a/include/gt-blas/bandsolver.h
+++ b/include/gt-blas/bandsolver.h
@@ -6,6 +6,10 @@
 
 #include "gt-blas/blas.h"
 
+#ifdef GTENSOR_DEVICE_SYCL
+#include <CL/sycl.hpp>
+#endif
+
 namespace gt
 {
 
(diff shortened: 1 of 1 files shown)


bd4:
gtensor relies pretty heavily on ifdefs already, I have not figured out a better way to do it. I guess one way to improve things, is to at least move things into different files, with `backend/sycl.h` and `backend/generic.h` and then use ifdef to conditionally include, that makes it a bit easier to read. That is how gt-fft and gt-blas are organized (they could also be improved with more clearly specifying the common interface that each backend must implement).

bd4:
Note that we are transitioning GENE to use gt-solver instead of gt-blas directly, so it would be useful to add this to gt-solver. I didn't bother adding a bandsolver type in gt-solver because all the platforms had a better alternative, dense is actually working a lot better on Intel now after some MKL optimization work.

bd4:
Finally got around to testing this, and there are matrix and bandwidth combinations where it is the fastest solver on certain hardware. I am incorporating this into a new PR that also adds `gt::solver::solver_band`, so it's easier to integrate into GENE with the new interface. Thanks again!

bd4:
Merged with #253, thanks!

PR #252, opened 2023-03-04
view: add view_strided, strided reshape etc overloads

bd4:
The data method raises an exception or assert(0) in device code if called on a non-forward-contiguous view.

### DIFF
--- a/include/gtensor/gview.h
+++ b/include/gtensor/gview.h
@@ -233,11 +233,7 @@ public:
   GT_INLINE decltype(auto) data_access(size_type i) const;
   GT_INLINE decltype(auto) data_access(size_type i);
 
-  inline pointer data() const;
-
   inline std::string typestr() const&;
-
-  inline bool is_f_contiguous() const;
 
 private:
   EC e_;
@@ -319,32 +315,12 @@ GT_INLINE decltype(auto) gview<EC, N>::data_access(size_t i)
 }
 
 template <typename EC, size_type N>
-inline typename gview<EC, N>::pointer gview<EC, N>::data() const
-{
-  if (!is_f_contiguous()) {
-#ifdef GTENSOR_DEVICE_ONLY
-    printf("cannot get data pointer for non-contiguous view");
-    assert(0);
-#else
-    throw std::runtime_error("cannot get data pointer for non-contiguous view");
-#endif
-  }
-  return pointer(&(data_access(0)));
-}
-
-template <typename EC, size_type N>
 inline std::string gview<EC, N>::typestr() const&
 {
   std::stringstream s;
   s << "v" << N << "(" << e_.typestr() << ")" << this->shape()
     << this->strides();
   return s.str();
-}
-
-template <typename EC, size_type N>
-inline bool gview<EC, N>::is_f_contiguous() const
-{
-  return this->strides() == calc_strides(this->shape());
 }
 
 template <typename EC, size_type N>
--- a/include/gtensor/gview.h
+++ b/include/gtensor/gview.h
@@ -117,6 +117,22 @@ struct select_reshape_gview_adaptor
   using type = detail::gview_adaptor<E>;
 };
 
+template <typename E>
+constexpr bool is_data_stride_expr()
+{
+  return is_gcontainer<E>::value || is_gtensor_span<E>::value;
+}
+
+/*
+template <typename E>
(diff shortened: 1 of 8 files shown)


tests/test_gtensor.cxx
```diff
@@ -951,6 +951,9 @@ TYPED_TEST(gtensor_copy, to_expr)
   auto b = gt::gtensor<double, 2, b_space_type>(a.shape(), 0.);
 
   gt::copy(a, b.view());
+  // Note: cuda allows device to device copy to be async, and the SYCL backend
+  // follows the convention
+  gt::synchronize();
   EXPECT_EQ(b, a);
```
germasch, line 957: But shouldn't the test `b == a` be a kernel, too, and hence correctly ordered? (Or, if the equality test happens on the cpu, it should involve a copy first, which would synchronize as well? Or maybe there's some subtlety involving managed memory?)

include/gtensor/gview.h
```diff
@@ -305,6 +325,12 @@ inline std::string gview<EC, N>::typestr() const&
   return s.str();
 }
 
+template <typename EC, size_type N>
+inline bool gview<EC, N>::is_f_contiguous() const
+{
+  return this->strides() == calc_strides(this->shape());
+}
+
```
germasch, line 333: Apparently, gview has a `strides()` method. Does it always? In particular, what happens if you a take a view of, say, `a + b`, where `a` and `b` have different strides?

germasch (commented):
I think the good news here are that this isn't going to break things. But I'm not necessarily convinced that it's the right direction to go, since it adds methods which can only be used on a subset of views. That may be okay, but it'd be nicer to follow something like, if the object this is a view of has a `data()` method, then the view has a `data()` method. Clearly, whether things are contiguous is something that only can be determined at runtime, though.

A minor thing: `f_contiguous` is from numpy and means Fortran (-order) contiguous.

One thing I've been considering for a while: If one takes a view of a strided thing, one should be able to do so by returning a new strided thing, with different strides but the same data() (possibly offset). This might also help to make things cheaper, or at least less complex, since a view of a gtensor_span really would just be another gtensor_span, possibly avoiding the chaining of objects (but there are issues with ownership if the underlying this owns data). Obviously, a gtensor_span already has data and strides, so that might mean that the case here might then automatically be covered.

bd4:
Special casing views that preserve strides is a good idea, but may be tricky to do, since the expression type is encoded in the return type, so for that to work, we have to know just from the TMP on the parameter types whether it needs to be a view or not. I think this may not be possible unless slice is encoded into a template param, which seems like a bad idea... since is contiguous when outermost is sliced, or when outermost dims are single index until a more inner slice.

Maybe the simplest thing, have a separate method and let programmer decide, like `gt::slice_f_contiguous(expr, slice)` that will do as much as possible with compile time checks and blow up an runtime if not contiguous, and always return span.

bd4:
It was easy to special reshape, flatten, swapaxis, and transpose. view was not so easy, because of the variadic template parameters and the auto return type, so I added `gt::view_strided` which only works on gtensor or gtensor_span and and returns a gtensor_span. I did not add a view_strided method to the containers - I think if this feature is needed, we can require the free function interface. There may be a clever way to specialize so it can use the same name, but is not easy.

I removed the gview data and is_f_contiguous, since the strided view returning gtensor_span should handle the use case of accessing data for fortran contiguous data, e.g. outer dim slices.

include/gtensor/gview.h
bd4: good point, I removed it form gview.

tests/test_gtensor.cxx
bd4: I re-tested and seems to be unnecessary, so I removed the commit for the PR. Not sure why it was failing without this at one point on Intel...

bd4:
This is still causing problems with some eval statements involving const. I think if rhs is const, then we can't pass through and need to force a copy / assign. I believe that the new behavior in gtensor_span is correct - shallow const means that if the top level object is const, it can still allow mutation (necessary for mutable lambdas to work). However if the element type is const, that should really not allow mutation. The to_kernel for const gtensor is even designed with this in mind, returning a const value_type span instead of just a const gtensor_span.

bd4:
I believe this is ready to merge.

bd4:
rebased

gmerlo (approved):

PR #256, opened 2023-03-17
cuda: add option to use rapids RMM pool allocator

bd4:
cmake option GTENSOR_USE_RMM


bd4:
Closing in favor of umpire PR including both with as much common code as possible

PR #260, opened 2023-04-03
use official cmake support for sycl and hip backends

bd4:


bd4:
Downside of this is that the SYCL cmake support requires cmake 3.23

germasch (approved):
This is definitely something I've been thinking we should do for quite a while ;)

bd4:
This break building GENE on Intel unfortunately, MKL related link error. So some more work will be required.

bd4:
The ROCm build does seem to work at least.

bd4:
On Intel systems / oneAPI versions where GENE fails to link, gtensor examples and benchmarks link and run just fine. Something about the GENE cmake is messing up the link chain, and it is not linking libmkl_sycl.

bd4:
Issue building GENE was because of options form included custom FindMKL.cmake conflicting with MKLConfig.cmake. Not too hard to workaround.

PR #267, opened 2023-05-27
add FFT support on host (via fftw)

germasch:
This PR makes it possible to use `gt::fft` even if actual GPU support is disabled. It's a prerequisite for building GENE with gtensor support on host-only. The goal isn't production-level support of host-only C++/gtensor GENE, but rather to simplify development of GENE on a local machine does not have GPUs -- because in regular host-only mode,  a lot of parts of GENE aren't even compiled, and obviously not tested. It might be useful for debugging, too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ if (NOT CMAKE_BUILD_TYPE)
 endif()
 
 include(cmake/CPM.cmake)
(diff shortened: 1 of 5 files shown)


bd4 (approved):
LGTM, works with clang-15 and gcc-13 on arch without drama. I did need to add cstdint to backend_host.h for clang-15, but that is unrelated to this PR.

PR #268, opened 2023-05-29
make gtsolver compile host-only

germasch:
It doesn't provide a sparse implementation -- but that's ok.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -35,8 +35,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+    - name: install packages
+      run: sudo apt-get update && sudo apt-get install -y libopenblas-dev libfftw3-dev liblapacke-dev
     - name: cmake
-      run: cmake -S . -B build-tests-${{ matrix.cxx }} -DGTENSOR_DEVICE=host -DCMAKE_BUILD_TYPE=RelWithDebInfo
+      run: cmake -S . -B build-tests-${{ matrix.cxx }} -DGTENSOR_DEVICE=host -DCMAKE_BUILD_TYPE=RelWithDebInfo -DGTENSOR_ENABLE_CLIB=ON -DGTENSOR_ENABLE_BLAS=OFF -DGTENSOR_ENABLE_FFT=ON -DGTENSOR_ENABLE_SOLVER=OFF
       env:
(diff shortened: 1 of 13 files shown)


germasch:
I guess a BTW comment: It looks like the CI runs every job twice, once triggered by "push", once by "pull_request". Not much of an issue, but I don't think it's supposed to be this way...

bd4:
Is the idea here to add a host backend for gt-blas, so the solver can actually run on host?

germasch:
> Is the idea here to add a host backend for gt-blas, so the solver can actually run on host?

Yes, well a host backend for gtsolver, since I had already done host for gtblas a while back. There's not much this actually adds, it doesn't add a CSR matrix / sparse solver, but all the other solvers (dense, invert, banded) now work on the host, too, since they generically rely on blas/lapack. (except for the npvt variant). (At least, the tests pass for me locally -- there's probably something that should be done in the CI to actually run these tests in a host-only config, though)

germasch:
Well, I tried adding blas/fft/solver to the host CI, but there's some OpenBLAS related issue, so I ended up doing just fft for the time being. I couldn't reproduce the blas issue on a local Ubuntu docker image, so well, pushing it off to later (unless @bd4 you're looking something not very fun to do...)

germasch:
Well, I think there is `crealf/cimagf`, but then the question would be, why doesn't openblas use that?

bd4 (changes requested):
cblas.get_handle test fails for me. Not sure if the test needs to be adapted, or if something else is wrong:
```
[ RUN      ] cblas.get_handle
/home/bda/fusion/gtensor/tests/test_cblas.cxx:290: Failure
Expected: (h2) != (h3), actual: 0x556f74dd2890 vs 0x556f74dd2890
```

This is with gcc 11 on ubuntu 22.04. The clang build issue seems to be a pre-existing issue, not related to changes in this PR.

bd4:
> Well, I think there is `crealf/cimagf`, but then the question would be, why doesn't openblas use that?

Seems like a clear bug in openblas, but I see no open issues about it, and would take ages for it to end up in HPC and LTS distros, so the obvious work around is to use our own macros instead of the openblas provided ones.

germasch:
LGTM -- thanks! I guess you already did all the requested changes yourself, right?

bd4 (approved):

PR #274, opened 2023-08-07
Pr/fix cuda warnings

bd4:


bd4:
This potentially breaks PSC: https://github.com/psc-code/psc/blob/main/src/include/psc_fields_cuda.h#L189

I am not sure how this is working in PSC though - the array version of adapt does a std::copy, which I do not expect to work. I think this could easily be replaced by positional argument version, which I believe is device safe.

germasch (approved):
I guess if it breaks PSC, I'll deal with it...

PR #275, opened 2023-08-09
Pr/cuda solver spsv

bd4:
Improves sparse solvers for CUDA and HIP backends.


bd4:
draft because this is built on top of #274

PR #279, opened 2023-09-22
workaround mem_info compile error on MacOS

germasch:
This doesn't make mem_info work correctly, but at least it fixes the build. Not sure it's worth it to fix this properly, though it could be done.


bd4 (approved):
Looks good. Sorry I never thought anyone would compile it on non Linux :P, I forget MacOS.

PR #285, opened 2023-12-01
Return device_count=0 in error case

td-mpcdf:
For me it seems to be more natural if the get_device_count function returns just 0 in the case of cudaErrorNoDevice and cudaErrorInsufficientDriver. The handling of this case should happen on the applications level, but at the moment the gtensor code aborts.

### DIFF
--- a/include/gtensor/backend_cuda.h
+++ b/include/gtensor/backend_cuda.h
@@ -180,7 +180,21 @@ public:
   static int device_get_count()
   {
     int device_count;
-    gtGpuCheck(cudaGetDeviceCount(&device_count));
+    cudaError_t code = cudaGetDeviceCount(&device_count);
+    switch (code) {
+      case cudaErrorNoDevice:
+        fprintf(stderr, "Error in cudaGetDeviceCount: %d (%s)\n", code,
+                cudaGetErrorString(code));
+        device_count = 0;
+        break;
+      case cudaErrorInsufficientDriver:
+        fprintf(stderr, "Error in cudaGetDeviceCount: %d (%s)\n", code,
+                cudaGetErrorString(code));
+        fprintf(stderr, "Did you start the job on a CPU partition?\n");
+        device_count = 0;
+        break;
+      case cudaSuccess: break;
+    }
     return device_count;
   }
 
--- a/include/gtensor/backend_cuda.h
+++ b/include/gtensor/backend_cuda.h
@@ -181,20 +181,17 @@ public:
   {
     int device_count;
     cudaError_t code = cudaGetDeviceCount(&device_count);
-    switch (code) {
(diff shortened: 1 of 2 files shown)


bd4:
I think this is a reasonable change and won't break anything, @germasch what do you think?

Regarding the implementation, you can use `gtGpuCheck(code)` to raise trigger standard error handling if the code is not success or no device (or anything else you want to special case). This would also need to be implemented for SYCL so behavior is consistent.

@td-mpcdf do you imagine writing code that would fallback to running on CPU if device count is 0?

include/gtensor/backend_cuda.h
```diff
@@ -180,7 +180,21 @@ class backend_ops<gt::space::cuda>
...
+                cudaGetErrorString(code));
+        device_count = 0;
+        break;
+      case cudaErrorInsufficientDriver:
+        fprintf(stderr, "Error in cudaGetDeviceCount: %d (%s)\n", code,
+                cudaGetErrorString(code));
+        fprintf(stderr, "Did you start the job on a CPU partition?\n");
+        device_count = 0;
+        break;
+      case cudaSuccess: break;
```
bd4, line 196: need a fallback case that calls `gtGpuCheck(code)` for all other errors.

include/gtensor/backend_hip.h
```diff
@@ -216,7 +216,14 @@ class backend_ops<gt::space::hip>
...
   {
     int device_count;
-    gtGpuCheck(hipGetDeviceCount(&device_count));
+    hipError_t code = hipGetDeviceCount(&device_count);
+    switch (code) {
+      case hipErrorNoDevice:
+        /* Set silently the return value to 0 */
+        device_count = 0;
+        break;
+      case hipSuccess: break;
```
bd4, line 225: Also needs fallback with `gtGpuCheck(code)`.

bd4:
Clang-format wants to put gtGpuCheck on one line, but that breaks the macro :P. I'm trying to figure out the least disruptive solution here. I am thinking maybe changing to an if/elseif/else clause.

bd4:
I think part of the problem is that gtensor dies on errors in get device count, which does not give the application a chance to interpret the error and decide whether it should die or do something else.

bd4:
This PR at least allows the application to raise some kind of helpful error to user if gpu count is 0 and it's trying to run in GPU mode. I think we merge this and can refine further in future. It could raise an exception or we could define gtensor error codes that standardize across the backends (hip and cuda are largely similar but SYCL tends to use exceptions).

bd4:
Regarding negative numbers as errors, I think that is reasonable, and we could have codes for negative numbers. However 0 for the special case of no device or no driver seems reasonable to me. My vote is to merge as is and refine in future PR.

PR #288, opened 2024-01-03
Add roctracer support

td-mpcdf:
- Load the right libraries in `CMakeLists.txt`
- Add call to roctracer API for `gpuProfilerStart` and `gpuProfilerEnd`


bd4 (approved):
Looks good thanks! Unfortunate that AMD does not provide cmake support for roctracer, I checked even in 5.5.0 and there are no relevant cmake files for e.g. doing find_package(roctracer), so I think hacking it using ROCM_PATH is the best we can do.

PR #290, opened 2024-09-16
Pr rocm6

gmerlo:
Minor fixes necessary for rocm>6

### DIFF
--- a/include/gt-solver/backend/hip-csrsm2.h
+++ b/include/gt-solver/backend/hip-csrsm2.h
@@ -9,7 +9,7 @@
 
 #include "gt-blas/blas.h"
 
-#include <rocsparse.h>
+#include <rocsparse/rocsparse.h>
 
 // ======================================================================
 // error handling helper
--- a/include/gtensor/backend_hip.h
+++ b/include/gtensor/backend_hip.h
@@ -260,9 +260,9 @@ public:
     }
     gtGpuCheck(rval);
 #if HIP_VERSION_MAJOR >= 6
-      return (attr.type == hipMemoryTypeDevice || attr.isManaged);
+    return (attr.type == hipMemoryTypeDevice || attr.isManaged);
 #else
-      return (attr.memoryType == hipMemoryTypeDevice || attr.isManaged);
+    return (attr.memoryType == hipMemoryTypeDevice || attr.isManaged);
 #endif
   }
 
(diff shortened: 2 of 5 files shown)


germasch (approved):
LGTM.

I'm not much of a fan of the `static_cast<void>`, but I suppose you got warnings (or worse) because of the ignored return value -- did you consider actually checking the return value? I usually just do something like

```
rc = hipFree(...);
assert(rc == 0) // or whatever the success return is....
```
Of course, this doesn't actually handle failure either...

src/fortran/gpu_api.cxx
```diff
@@ -88,7 +88,7 @@ extern "C" void gpuDeviceReset()
 #ifdef GTENSOR_DEVICE_CUDA
   cudaDeviceReset();
 #elif defined(GTENSOR_DEVICE_HIP)
-  hipDeviceReset();
```
bd4, line 91: This should probably be gtCheck, we don't really want to ingore the return value.

include/gtensor/backend_hip.h
```diff
@@ -268,19 +272,27 @@ class backend_ops<gt::space::hip>
     hipPointerAttribute_t attr;
     auto rc = hipPointerGetAttributes(&attr, ptr);
     if (rc == hipErrorInvalidValue) {
-      hipGetLastError(); // clear the error
+      static_cast<void>(hipGetLastError()); // clear the error
```
bd4, line 275: Here we are purposely ignoring the return val to clear the error. I think C style (void) is easier to read, but don't have a strong opinion.

include/gt-solver/backend/hip-csrsm2.h
```diff
@@ -9,7 +9,7 @@
 
 #include "gt-blas/blas.h"
 
-#include <rocsparse.h>
+#include <rocsparse/rocsparse.h>
```
bd4, line 12: does this still work with Rocm 5.x? I think we can drop support for 4.x now, but not sure we want to drop 5 yet.

bd4 (changes requested):

bd4:
CI failure is unrelated, I guess we were pulling dev branch of umpire and it wants newer cmake.

gmerlo:
this was me mostly being lazy and avoiding warnings. There is however a major regression with rocm6, as if the memory is not correctly allocated anymore. But could also be something else, not sure yet what is going so I'll keep this a draft for a little longer

bd4:
merged same feature with backwards compat for <5.5 in #292

PR #292, opened 2024-09-19
Step to newer rocm versions and vendor independent naming

td-mpcdf:
This PR
- adapts to the new include directory structure of ROCM (without giving up the old one)
- changes a CMAKE variable name from CUDA to general GTENSOR_GPU_ARCHITECTURES naming
-


bd4:
Have you tested this with ROCM 5.5 and < 5.5? The include change happened in 5.5. I tried changing the ifs to check for 5.5 and later, and it still fails to compile tests even with `-Wno-error`, because google tests insists on compiling with `-Werror` which includes the gtensor tests that have a bunch of rocm warnings. The way Werror is handled in C++ is kind of garbage, breaks things with new releases and causes major headaches. I think even if we use correct new includes, ROCm's own includes in 5.5 have internal deprecated includes triggering warnings.

Given that GENE is probably the only user of this, do we need to support ROCM <5.5 or even <6? If we do, then I want to make sure we can verify it works, otherwise we can just assume the new include paths and not worry about backward compat.

bd4 (approved):
Looks good, thank you for the contribution!

We can also consider removing the <5.5. compatibility in the future.

PR #296, opened 2025-11-13
use C++ std17 always

gmerlo:
Update to  c++17 always as is needed for rocm/7 (rocmprim)


bd4:
CUDA still only needs 14 I think, but I have no objection to always requiring 17 - that would also allow us to simplify some code going forward.

PR #300, opened 2026-03-02
Add first-class HIP support

td-mpcdf:
Since CMake 3.21, the HIP language is supported. This PR enables the hip language in the hip case.
Also updated CPM and GTest to more recent versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,7 +166,7 @@ if ("cuda" IN_LIST GTENSOR_BUILD_DEVICES)
     CPMAddPackage(
       NAME rmm
       GITHUB_REPOSITORY rapidsai/rmm
-      VERSION 23.02.00
+      VERSION 26.02.00
     )
     target_compile_definitions(gtensor_cuda INTERFACE
       GTENSOR_USE_MEMORY_POOL GTENSOR_USE_RMM)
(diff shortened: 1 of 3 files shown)


bd4:
Yes and once I updated the cmake install, googletest failed to build. I think the CI is overdue for some verison bumps anyway, there is no reason to use 6 year old images for this project. CI can be a real pain because it's not easy to test locally - I can work through the issues and get something working, lets not duplicate effort.

bd4:
can you rebase off of the latest main branch? You can use `git rebase -i origin/main`  and just remove the old ci fix commit and it should work.

td-mpcdf:
Still a problem that I do not understand. Any idea?

bd4:
If nobody is using RMM, we could remove from CI and deprecate. Looks like you are on the right track for fixing it though.

td-mpcdf:
I didn't know about RMM, but it looks interesting, so I will try it in the future and would keep it in the CI. But with the upgraded version 26.0.2.00 it needs cmake>3.30.4. You only upgraded to 3.21. Is there any problem in going to the last of the major version 3?

td-mpcdf:
The updated RMM 26.02.00 has a changed API, so it does not work with code changes and I am not familiar with the RMM part. The old RMM 23.02.00 does somehow not provide rmm::rmm target for CMake anymore???

bd4:
> The updated RMM 26.02.00 has a changed API, so it does not work with code changes and I am not familiar with the RMM part. The old RMM 23.02.00 does somehow not provide rmm::rmm target for CMake anymore???

I haven't looked at RMM in a while so I am not sure what we are supposed to change re the targets.

Re cmake version, I am open to bumping it further... 3.21.0 was released in 2021. Ubuntu 22.04 has 3.22.1, and 24.05 has 3.28.3. So if we go to minimum 3.20, we are requiring third party repos or install from source. I don't love having that requirement on the latest LTS, I am more comfortable with 3.28 as minimum. But we could just set 3.28 or 22 in the cmake file and use 3.30 in the CI for the RMM pipeline only.

PR #301, opened 2026-03-02
ci: use newer cuda image and cmake 3.21

bd4:


bd4:
@td-mpcdf I think the permissions allow anyone to do a review?

PR #303, opened 2026-03-05
Pr/cuda13

td-mpcdf:
Add a test for cuda13 and adapt backend_cuda to also work with cuda13 (small API changes).


bd4 (approved):
Looks good, thanks! Just waiting on CI after merging main branch with the hip pr

PR #304, opened 2026-03-05
Add native hip support

td-mpcdf:
Now with minimal changes to enable HIP support from CMake.
I also changed the check for the CXX compiler to be hipcc or hcc. In the amd/rocm case, amdclang++ is also a valid CXX compiler. In general the CMAKE_HIP_COMPILER can be different from the CMAKE_CXX_COMPILER.


germasch (approved):
Thanks, looks great to me!

PR #305, opened 2026-09-22
Pr uneven ranks

gmerlo:
Allow for uneven number of ranks per node


bd4:
Can you rebase on laster main branch? Otherwise this looks fine to me. Would be nice to have a SYCL implementation, but if it's not being actively used at the moment and SYCL test still passes, I guess is fine for now.

